PR #2, opened 2018-04-15
Additional tests and changes to trouble random

VanOvermeire:
Hi,

I like the idea of creating a library for introducing chaos monkey in Spring Boot (might be able to use this at work), so decided to contribute a little bit. 

Summarized, the pull request does the following:

- tries to improve test coverage by adding some unit tests
- does some cleanup (like removing exceptions that aren't thrown)
- reworks the troubleRandom

Last part probably requires some explanation. The issue I have with getTroubleRandom() is that it does not match the documentation, which says:

> "chaos.monkey.assaults.level: How many requests are to be attacked. 1 each request, 5 each 5th request is attacked"

Which I think means when the level is 1, **every** request is attacked.

But this:

`return RandomUtils.nextInt(0, 10);`

Will return numbers **including** 0, meaning at level 1, the condition will sometimes be false and chaos monkey will ignore some 10% of requests. And if you fill in 10, chaos monkey will not attack every 10th request. Instead, it will never do anything (RandomUtils is endExclusive).

You could change the documentation, but I think it is more intuitive when level 1 means every request is attacked. So I changed the code.

In case you think otherwise (or if the above thought process here is wrong), I separated that change from the others, putting it in a different commit, so you can ignore it more easily. :)

Thanks

### DIFF
--- a/spring-boot-chaos-monkey/pom.xml
+++ b/spring-boot-chaos-monkey/pom.xml
@@ -55,6 +55,11 @@
             <version>3.4</version>
         </dependency>
         <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+            <version>${spring-boot.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
             <version>${jcl-over-slf4j.version}</version>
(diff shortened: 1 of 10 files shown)


MrBW:
Thanks yo so much for this pr, I will merge it as it is. Please feel free to share your thoughts and please feel free to contribute. Currently I´m preparing a metric endpoint and a configuration endpoint. If you have any ideas, your are welcome!

PR #4, opened 2018-04-19
chaos.monkey.watcher.services -> chaos.monkey.watcher.service

wonwoo:
is not this a bug?

WatcherProperties.class
```java
    @Value("${service:true}")
    private boolean service;
```
AttackServiceCondition.class

```java
.getProperty("chaos.monkey.watcher.services","false")
```


MrBW:
Yes! Thx for spotting me on... ;-)

Issue #5, opened 2018-04-19
Add support for Spring Boot 1.5.x


cgew85:
Since a lot of projects are still based on earlier versions of Spring Boot, adding support for them might be a good idea.

MrBW:
Hi,

implementation for Spring 1.5.x is done...
I´ve created a first SNAPSHOT version for you, please take a look at:
https://oss.sonatype.org/content/repositories/snapshots/

<dependency>
  <groupId>de.codecentric</groupId>
  <artifactId>chaos-monkey-spring-boot</artifactId>
  <version>1.5.0-SNAPSHOT</version>
</dependency>

Documentation can be found here: 
https://codecentric.github.io/chaos-monkey-spring-boot/1.5.0-SNAPSHOT/

Bye,
Benjamin

PR #6, opened 2018-04-20
Spring boot 1.5.x

wonwoo:
what do you think? 

#5 spring boot 1.5.x 

think about it.

### DIFF
--- a/spring-boot-chaos-monkey/pom.xml
+++ b/spring-boot-chaos-monkey/pom.xml
@@ -12,10 +12,10 @@
     <name>Chaos Monkey for Spring Boot</name>
     <url>https://github.com/codecentric/chaos-monkey-spring-boot</url>
     <properties>
-        <revision>1.0.2-SNAPSHOT</revision>
-        <spring.version>5.0.5.RELEASE</spring.version>
+        <revision>1.0.3-SNAPSHOT</revision>
+        <spring.version>4.3.16.RELEASE</spring.version>
         <jcl-over-slf4j.version>1.7.25</jcl-over-slf4j.version>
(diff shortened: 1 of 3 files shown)


MrBW:
I´m not sure how to handle it...maybe we could support both. All new ideas and improvements can be supported by Chaos Monkey for Spring Boot 2.x. Let me think about it....because there are so many customers on Spring Boot 1.5.x...

What do you think, support both or not?

MrBW:
Sounds great! How we should handle it? Should we start a branch for Spring Boot 1.5, because I want to attack more Components, which are only available in Spring Boot 2.

We could split our versions, followed by the Spring Boot version:
- chaos-monkey-spring-boot 1.0.x
- chaos-monkey-spring-boot 2.0.x

Any other ideas?

tiarebalbi:
I think that's it, having two version will be a good way to keep up-to-date with the changes that are happening with spring 2.x and keep the old version working. 

The approach, I would go with the option of having a branch for 1.5 which will allow us to cherry-pick any of changes required or needed in that branch.

MrBW:
Ok! Starting by 1.0.2 or maybe with 1.5.x for our Spring 1.5 support?

tiarebalbi:
I would go 1.5.x.

MrBW:
Ok, we will start a 1.5.x branch as soon as we have add some new features. We need to provide a dynamic configuration and metrics for both 1.5.x and 2.0.x. https://github.com/chaostoolkit would like to integrate the Chaos Monkey for Spring in there Chaos Engineering tool suite.

I will close this pr, because we can´t start at the moment and have to add the mentioned features.
Thanks for your PR and keep contributing, you are welcome! :-)

PR #11, opened 2018-04-25
Replacing profile to property

tiarebalbi:
Changes to replace the `@ProfileActive` with application property. Let me know your thoughts on this.

### DIFF
--- a/chaos-monkey-demo-app/src/main/resources/application.properties
+++ b/chaos-monkey-demo-app/src/main/resources/application.properties
@@ -1,4 +1,4 @@
-spring.profiles.active=chaos-monkey
+chaos.monkey.enabled=true
 chaos.monkey.assaults.latencyActive=true
 chaos.monkey.watcher.controller=true
 chaos.monkey.assaults.level=3
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
@@ -11,6 +11,7 @@ import de.codecentric.spring.boot.chaos.monkey.watcher.SpringRestControllerAspec
 import de.codecentric.spring.boot.chaos.monkey.watcher.SpringServiceAspect;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.*;
 import org.springframework.core.io.ClassPathResource;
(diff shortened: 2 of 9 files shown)


MrBW:
Looks good....but if we changed to dynamic configuration, what happened with the @Configuration class. Do we need a Application Context refresh to trigger the init of the Chaos Monkey? I´m not totally sure if the @Configuration class will be initialized on a property change at runtime.

tiarebalbi:
It's a good point, I was investigating it here, we could have different ways to disable or enable it. What do you think to have an `@WriteOperation` at the `ChaosMonkeyEndpoint.class` that would enable or disable the services? *(I still investigating other options)* 

To update the settings we could unregister and register the beans again, but this would be a manual process, the other option that I see would write an operation method that will change the settings (works with Spring Cloud Config as well) and at the moment we call the `ChaosMonkey#callChaosMonkey` we could check if the setting is enabled or not.

What do you think? 

**ChaosMonkeyEndpoint.class**
```
    @WriteOperation
    public void updateChaosMonkeySettings(@Selector Boolean enabled) {
        this.chaosMonkeySettings.getChaosMonkeyProperties().setEnabled(enabled);
    }
 ```
**ChaosMonkey.class**
```
   public void callChaosMonkey() {
        if (isTrouble() && isEnabled()) {
        .....
        }
    }
```

The `@ConditionalOnProperty` won't have the effect of enabling and disabling the beans as we are initializing all the beans in the bootstrap of the application.

**Comment by MrBW** [comment]
Looks quite awesome!

One thing I´m just thinking about...shall we stay with **@Profile** in our **@Configuration**? Why? Because if we start a Chaos Engineering Experiment, we have to control the behave of the monkey. Let me explain...

You start a new experiment and you have pick up a service you would like to infect with the CM. 

- First step, enable the CM by adding the profile "chaos-monkey" and start the app. 
- Second step, CM is loaded and everything is prepared. Now you have to enable the CM and start your experiment. As soon as you see something went out of control and you break your blast-radius for this experiment, you have to stop the CM immediately!

What do you think, should we stay on @Profile and implement your changes in the ChaosMonkey class? I would like to merge your changes in ChaosMonkey class, but before let us discuss a additional property or just stay on @Profile.

**Comment by tiarebalbi** [comment]
I see your point, I think we could go with the profile but just thinking here, if we go ahead with profile we would need the property anyway to mark as enabled or not.  As far as I remember, to change the profile we would have to restart the app to set the new profile right? 

In the other hand if we keep the property we could have the extra properties and remove the `@ConditionalOnProperty` from ChaosMonkeyConfiguration.class since execution will happen only if the `chaos.monkey.enabled == true`. In case of a break they just need send a POST to the `/actuactor/chaosmonkey?enabled=false`

Let me know what do you think, I'm in for both approach, once we are happy with the solution I will update the PR and prepare the changes for the 1.5.x (I think this could be a point to start the 1.5.x branch right?)

I tested the example above and I was able to see the new property right away, for example, using the http://localhost:8080/actuator/configprops

screen shot 2018-04-26 at 7 56 25 am

**Comment by MrBW** [comment]
We have to stay on a @ConditionalOnProperty or @Profile, because all AOP Watcher are initialized and will scan the infected app. I would like to stay on the @Profile and add your chaos.monkey.enabled in ChaosMonkey class.

So, feel free to update your PR and we can merge it ;-)

Starting 1.5.x branch makes sense, could we implement a basic RestController to update the configuration, so we could use it in 2.0.x and 1.5.x?

**Comment by tiarebalbi** [comment]
You’re right I just thought about it, was on my way edit the comment :D yes we can use the same rest controller for both. Preparing the changes !

File: chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/controller/ChaosMonkeyController.java [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+@RestController
+public class ChaosMonkeyController {
+
+    private final ChaosMonkeySettings chaosMonkeySettings;
+
+    public ChaosMonkeyController(ChaosMonkeySettings chaosMonkeySettings) {
+        this.chaosMonkeySettings = chaosMonkeySettings;
+    }
+
+    @PostMapping("/chaos-monkey/setting")
```
tiarebalbi: @MrBW Not really sure about this name here... What do you think?

MrBW: Hmmm....what about:

-settings (it´s more than 1)
-configuration
-properties

I´m fine with all of them! ;-)

MrBW:
Ready to merge?

MrBW:
Thank you so much!!!

tiarebalbi:
Just let me know if you have any other ticket that I could be picking!

Thanks

MrBW:
I found a bug, because now we are attacking our own Controller :-) already fix it and I´m preparing a PR. If you want I will assign you the PR to review it?

tiarebalbi:
Ops! Sure!! You can me to all of them ! Thanks

Issue #27, opened 2018-06-29
Assault level max


geramirez:
We love this library! Would you mind if we submitted a pr to raise the assult level max from 10 to 1000? Thanks!

geramirez:
Yep our service is hit at a rate of 900 request per second. We would like to test our availability at 4 9th to 5 9th. So we need Chaos Monkey to be much less active. 

We are running all the experiments.

PR #32, opened 2018-07-05
Bumped up max level to 100 for scenarios requiring sub-10% chaos

mattk-pio:
This is a PR for issue #27 ; Tried to keep it as minimal and no-impact to anyone currently using the project. I think going forward it might feel a bit more intuitive to switch to percentages (where 100 is always, as opposed to 1), but it's not a big deal.

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/configuration.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/configuration.adoc
@@ -17,7 +17,7 @@ Except for the Watcher, you can also influence the behavior of the Chaos Monkey 
 |chaos.monkey.assaults.level
 |How many requests are to be attacked.
  1 each request, 5 each 5th request is attacked
-|1-100
+|1-1000
 |5
 
 |chaos.monkey.assaults.latencyRangeStart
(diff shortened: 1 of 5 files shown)


MrBW:
Thx for your PR, I will come back on this...

miklosbarabas:
@MrBW one question: shouldn't the `getTroubleRandom` method use the `level` field as the upper bound of the range of the pseudorandom being returned? Like:
`public int getTroubleRandom() {
        return RandomUtils.nextInt(1, level + 1);
}`

deelthor:
@miklosbarabas interesting question. At least the behavior would be a bit different. imagine a level of 10. `RandomUtils.nextInt(1, level + 1)` will always return an int between 1 and 11. But the ChaosMonkey will only be executed when `chaosMonkeySettings.getAssaultProperties().getTroubleRandom() >= chaosMonkeySettings.getAssaultProperties().getLevel()`. So a level of 10 would mean a 10% chance, 100 a 1% chance etc., right? I am not sure whether this is the intended behavior. @MrBW has to decide.

miklosbarabas:
@deelthor right, at least that is what I understood from the code. Btw the documentation is a bit misleading I think as it is stating: 
> How many requests are to be attacked. 1 each request, 5 each 5th request is attacked.
Which is not entirely true, it's more like what you said: a level of 10 means a 10% chance for the assault to happen (even previously before this MR). I might be mistaken tho.

PR #45, opened 2018-09-07
Fix trouble calculation

snieking:
I noticed that the trouble calculation was inaccurate while running.

If you for example configured a level of 10, the isTrouble would return a value in the range of 1-1000 which does not mean every 10th request.

Instead I rewrote it to use the configured level(+1) as the upper bound of the range. And after this, the isTrouble calculation is working a lot better.

This new way of calculating trouble will also allow for a more flexible level configuration. I increased the max level to 10000 now, but this could potentially even be increased to max int, if that would be wanted.

I also moved isEnabled() before isTrouble(), so that we do not need to retrieve randoms all the time when chaos monkey is disabled.

---

Also, not a biggie, but during my last contribution I wasn't added as a contributor on the project. I am not sure how that works, and if there are any other requirements to become a contributor other than having a commit merged in.

### DIFF
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkey.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkey.java
@@ -45,7 +45,7 @@ public class ChaosMonkey {
     }
 
     public void callChaosMonkey(String simpleName) {
-        if (isTrouble() && isEnabled()) {
+        if (isEnabled() && isTrouble()) {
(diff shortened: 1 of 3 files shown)


MrBW:
Again, you are not on the contributor page...why? What I'm doing wrong? Merge via mobile phone is not working, maybe?

snieking:
Hehe possibly, thanks for accepting the PR! 👍

MrBW:
Could you please send a dummy PR to the master branch? I changed the protection rules on this branch. Maybe it was something about this, because in the past it worked as usual...

snieking:
Hi, from a fork or a branch on this project? I am getting permission denied on the latter.

MrBW:
PR from a fork please

PR #46, opened 2018-09-10
Update documentation to reflect the new increased max level

snieking:


MrBW:
Ah...ha! Secured master was the problem...! Welcome aboard! ;-)

image

snieking:
There we go, thanks!

Issue #86, opened 2019-08-05
Chaos Monkey for Spring boot not working with java11


maiksensi:
Seems chaos-monkey-for-spring-boot is currently not working with java11. We should change that at some point. (Guess I would mark this as a feature request)

Basically two tests are failing:
```
Results :

Failed tests:   allowInterruptionOfAssaultDuringHoldPeriod(de.codecentric.spring.boot.chaos.monkey.assaults.MemoryAssaultIntegrationTest): Memory did not reach half exhaustion before timeout

Tests in error: 
  initializationError(de.codecentric.spring.boot.chaos.monkey.component.ChaosMonkeySchedulerTest): java.lang.reflect.InvocationTargetException
  initializationError(de.codecentric.spring.boot.chaos.monkey.assaults.LatencyAssaultTest): java.lang.reflect.InvocationTargetException

Tests run: 87, Failures: 1, Errors: 2, Skipped: 0
```

How to reproduce: 
- use java11 locally and run `mvn test`

WtfJoke:
The InvocationTargetException has something to do with the used PowerMock/Mockito-Version (probably the one used is not java9+ compatible)

WtfJoke:
Closed by #87

PR #87, opened 2019-08-09
Make tests run with Java 12 (by getting rid of powermock)

WtfJoke:
Hi there

So this Pull-Request (inspired by https://github.com/codecentric/chaos-monkey-spring-boot/issues/86) does the following things:

- :heavy_minus_sign: Removing powermock(ito) as dependency
- :heavy_minus_sign: :arrow_up: Remove mockito_all ([not produced anymore](url)) with mockito-core and upgraded to latest version (from 1.10 to 3.0)
- :sparkles: Introduced a ChaosMonkeyLatencyAssaultExecutor to get the test LatencyAssaultTest green
- :green_heart: Make Travis run on jdk9-12
- :white_check_mark: :fire: Removed unnecessary stubs (I figured out, that KillAssault is currently not tested in ChaosMonkeyRequestScopeTest and probably should be tested in a new ChaosMonkeyRuntimeScopeTest)

Up so far this PR does not fix the flanky testcase MemoryAssaultIntegrationTest#allowInterruptionOfAssaultDuringHoldPeriod, which sometimes fails.

I would like to have some feedback in following subjects:

1. Introducion of ChaosMonkeyLatencyAssaultExecutor (I dont like the solution that much, if you have a better idea, how to solve it in a cleaner/better way, let me know)
2. I needed to add mockito-core as dependency on the subprojects chaos-monkey-demo-app-naked and chaos-monkey-demo-app. If I remove it we get some classnotfoundexceptions from spring mocking classes (I dont know why thats the case)

~~3. On my local (windows) machine, the test ChaosDemoApplicationTests#checkMetricsBean (of Project chaos-monkey-demo-app-ext-jar only) is failing. On travis (linux) its no problem. Is that the case for you too? Do you have an idea, why the metrics bean could be not available?~~ It fails on current master as well (created #90).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: java
-jdk: openjdk8
+jdk:
+  - openjdk8
+  - openjdk9
+  - openjdk10
+  - openjdk11
+  - openjdk12
 sudo: false # faster builds
 
 after_success:
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/ChaosMonkeyLatencyAssaultExecutor.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/ChaosMonkeyLatencyAssaultExecutor.java
@@ -1,0 +1,6 @@
+package de.codecentric.spring.boot.chaos.monkey.assaults;
+
+public interface ChaosMonkeyLatencyAssaultExecutor {
+
+    void execute(long duration);
+}
--- a/chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScopeTest.java
+++ b/chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScopeTest.java
@@ -143,7 +143,6 @@ public class ChaosMonkeyRequestScopeTest {
     public void isExceptionActiveExpectExceptionAttack() {
         given(exceptionAssault.isActive()).willReturn(true);
         given(latencyAssault.isActive()).willReturn(false);
-        given(this.assaultProperties.chooseAssault(2)).willReturn(0);
 
         chaosMonkeyRequestScope.callChaosMonkey(null);
 
@@ -155,7 +154,6 @@ public class ChaosMonkeyRequestScopeTest {
     public void isLatencyActiveExpectLatencyAttack() {
         given(exceptionAssault.isActive()).willReturn(false);
         given(latencyAssault.isActive()).willReturn(true);
-        given(this.assaultProperties.chooseAssault(2)).willReturn(0);
 
         chaosMonkeyRequestScope.callChaosMonkey(null);
 
@@ -196,7 +194,6 @@ public class ChaosMonkeyRequestScopeTest {
     public void chaosMonkeyIsNotCalledWhenServiceNotWatched() {
         String customService = "CustomService";
 
-        given(exceptionAssault.isActive()).willReturn(true);
         given(this.assaultProperties.getWatchedCustomServices()).willReturn(Collections.singletonList(customService));
         given(chaosMonkeySettings.getAssaultProperties().isWatchedCustomServicesActive()).willReturn(true);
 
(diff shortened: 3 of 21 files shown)


chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/ChaosMonkeyLatencyAssaultExecutor.java
```diff
@@ -0,0 +1,6 @@
+package de.codecentric.spring.boot.chaos.monkey.assaults;
+
+public interface ChaosMonkeyLatencyAssaultExecutor {
```
maiksensi: I think the introduction of an interface for assaults makes sense. Should we rename this to `ChaosMonkeyAssaultExecutor`? After all I guess an interface makes sense if it is not bound to a specific assault right? Should we then start refactoring all assaults to implement that interface for their specific `execute`s? I am not that firm with the codebase so far, but to me it looks like this shouldnt be a specific assault interface.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScopeTest.java
```diff
@@ -207,31 +190,28 @@ public void isChaosMonkeyExecutionDisabled() {
...
-        verify(killAppAssault, never()).attack();
     }
 
     @Test
     public void chaosMonkeyIsNotCalledWhenServiceNotWatched() {
         String customService = "CustomService";
 
         given(exceptionAssault.isActive()).willReturn(true);
-        given(this.assaultProperties.getWatchedCustomServices()).willReturn(Arrays.asList(customService));
+        given(this.assaultProperties.getWatchedCustomServices()).willReturn(Collections.singletonList(customService));
```
maiksensi: ❤️

.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: java
-jdk: openjdk8
+jdk:
+  - openjdk8
+  - openjdk9
+  - openjdk10
+  - openjdk11
+  - openjdk12
```
maiksensi: 👍 🥇 ❤️

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/ChaosMonkeyLatencyAssaultExecutor.java
WtfJoke: Yes, thats a good idea. In the beginning I had a similar idea, but thought it could be confused with the Interface ChaosMonkeyAssault. 

As far as I understand all Aussaults are split between ChaosMonkeyRequestAssault and ChaosMonkeyRuntimeAssault. Where as the Interface ChaosMonkeyAssault is a legacy one (and shouldnt be used anymore, according to javadoc)

The question remain, what will be the purpose of the ChaosMonkeyRequestAssault/ChaosMonkeyRuntimeAssault if we move the attack logic into separate executors? Are they just for metrics, stuff and the call of the executor?

WtfJoke:
I think dont like that I just have another interface and another class, which I need pass as an additional parameter into the AssaultClass (Just for the purpose of mocking it away). 

I can live with it, but maybe theres a better way out there, I cant see.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/ChaosMonkeyLatencyAssaultExecutor.java
maiksensi: @MrBW can you give us a hint, whether the naming is fine like it is or whether going for a general `ChaosMonkeyAssaultExecutor` makes sense. In that case I guess we would gradually refactor and remove the other interfaces like `ChaosMonkeyRequestAssault`. Or maybe a short explanation why they are there makes sense, I guess I just don't get the idea so far.

maiksensi:
> I think dont like that I just have another interface and another class, which I need pass as an additional parameter into the AssaultClass (Just for the purpose of mocking it away).
> 
> I can live with it, but maybe theres a better way out there, I cant see.

I don't think it is only for mocking? It also hides the "implementation" details of the executor away, which makes the module and tests more robust in the end.

PR #95, opened 2019-08-27
Update Snapshot to correct milestone

maiksensi:


pom.xml
```diff
@@ -29,13 +30,14 @@
 
     <properties>
         <!-- project version -->
-        <revision>2.0.3-SNAPSHOT</revision>
+        <revision>2.1.0-SNAPSHOT</revision>
```
WtfJoke: Only this line should change.

maiksensi: the file was not formatted properly. If you want I can create a separate MR, but the changes are in format only, so I actually count that as trivial and don't think this should be done in another MR.

maiksensi: as discussed lets make two commits.

WtfJoke:
With 2 commits its great :) Thanks! 👍

PR #96, opened 2019-08-29
Refactor / Fix Memory allocation

### LINKED ISSUE #92: MemoryAssault not working on systems with more than 32GB RAM
## Expected Behavior
When setting up a memory assault I expect it to work in the following:
- Docs tell me what I need to consider when using this type of attack (like that there is a grain of error for such attacks, and that they need to be fine-tuned from system to system). 

## Current Behavior
- On systems with more RAM than 32GB memory assault does not work propery (available heap and memory fill percent do not correlate as expected)

## Possible Solution
- Refactor the corresponding attack to be more concise and easier to understand. Refactor it so that it does one thing and one thing only.
- Add docs so with good examples so that it is clear to a possible user on what to expect from that attack. 

## Steps to Reproduce
- Use current snapshot build
- Run for example on java 11 >= 32GB RAM
- run `MemoryAssaultIntegrationTests`

WtfJoke:
Refactor memory assault
Fix (hopefully) flanky memory assault test (on all jdks)

Fixes #92


chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/MemoryAssault.java
```diff
@@ -141,4 +161,15 @@ private void waitUntil(int ms) {
...
         }
     }
+
+    /**
+     * Java 8 or lower: 1.6.0_23, 1.7.0, 1.7.0_80, 1.8.0_222
+     * Java 9 or later: 9.0.1, 11.0.4, 12, 12.0.1
+     *
+     * @return true if Java <= 8
+     */
+    private static boolean isJava8OrLower() {
```
maiksensi: I like that you moved that to a method, even tough I don't see the need for it. What I don't like is that you increased the matching range, even tough wer're not supporting java versions below 8. So I would suggest keep the method (maybe rename to `isJavaVersion8()`) and match specifically for 1.8 java versions.

WtfJoke: I am going to change it to Java 8 only. But in this case I preferr a variable again, so I will change it (but only local scope of the method)

WtfJoke: I just changed it, check it out 👍

PR #104, opened 2019-09-02
Deploy snapshots

### LINKED ISSUE #98: Automatic snapshot generation
snapshots, docs (including changelog) should be generated automatically when merging to master.

WtfJoke:
Will deploy each snapshot produced by build on master to OSSRH
Fixes #98

### DIFF
--- a/.travis.settings.xml
+++ b/.travis.settings.xml
@@ -1,0 +1,12 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                      http://maven.apache.org/xsd/settings-1.0.0.xsd">
+  <servers>
+    <server>
+      <id>ossrh</id>
+	  <username>${env.SONATYPE_USERTOKEN}</username>
+      <password>${env.SONATYPE_USERTOKENPASSWORD}</password>
+    </server>
+  </servers>
+</settings>
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,22 @@ jdk:
   - openjdk10
   - openjdk11
   - openjdk12
-sudo: false # faster builds
-
+cache:
+  directories:
+    - $HOME/.m2
 after_success:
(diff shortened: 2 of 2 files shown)


maiksensi:
Hey @WtfJoke thanks for the contribution :). 
I think the snapshot generation and the docs update should be tackled together.
What do you think about also adding the docs generation to this MR?

Is it me or is the signing part missing? If so, I guess we should add that in any case, so that each artifact is signed.

WtfJoke:
Hi @maiksensi 

I am not really sure how to generate the docs, if you tell me how to achieve that I can implement it.  :D

Youre right, signing is not there, snapshots doesnt need to be signed. If we implement release signing, we can also use that to sign snapshots (but this would exceed the size of this PR and I would preferr to to that in the automatic release PR).

maiksensi:
> Hi @maiksensi
> 
> I am not really sure how to generate the docs, if you tell me how to achieve that I can implement it. :D
> 
> Youre right, signing is not there, snapshots doesnt need to be signed. If we implement release signing, we can also use that to sign snapshots (but this would exceed the size of this PR and I would preferr to to that in the automatic release PR).

Generating the docs:
* switch directory to `chaos-monkey-docs` 
* run `mvn generate-resources` which will render the docs under ./target/generated-docs/
*  merge the the content of that folder to the correct version folder in branch `gh-pages` 
--> For version 2.1.0-SNAPSHOT: `./images` and `index.html` need to be merged to branch `gh-pages` folder `./2.1.0-SNAPSHOT/images/` and `./2.1.0-SNAPSHOT/index.html`
* Version is the same as the one setup in the main `pom.xml`. 

Youre right, we can take care of signing and stuff in a later MR.

WtfJoke:
> what happens if the `pom.xml` version is set to, let's say 2.1.0 instead of 2.1.0-SNAPSHOT? I that also deployed as a snapshot with that version number?

I would deploy it as a full release (as soon as the signing part is also done) to maven 😅

WtfJoke:
I would merge it like that 👍  (What im not sure is, if it would try to deploy two times on a non snapshot release or just once)

Need new approval, since there are new commits @maiksensi (Im sorry 🍹 🌴 )

PR #116, opened 2019-09-06
Fix version reference

WtfJoke:
**What**:
Fixing reference to old version in README of gh-pages

**Why**:
We shouldnt reference to specific versions (or in this case old versions)

**How**:
Replaced old version 2.0.1 with X.X.X

**Checklist**:

- [ ] Documentation added
N/A
- [ ] Changelog updated
N/A
- [ ] Tests added
N/A
      
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Let´s activate Chaos Monkey for Spring Boot, only 2 steps are required.
 <dependency>
     <groupId>de.codecentric</groupId>
     <artifactId>chaos-monkey-spring-boot</artifactId>
-    <version>2.0.1</version>
+    <version>X.X.X</version>
 </dependency>
 {% endhighlight %}
 2. Start your app enabling the service using the property below:


maiksensi:
Hey @WtfJoke can you explain a bit, what does improves? It seems to me the change is rather confusing if users copy & paste the `pom` entry. I would either always use the newest version here or add a note that X.X.X needs to be substituted for the newest CM version.

What do you think?

WtfJoke:
It improves that no old chaosmonkey will be used, just by copy pasting from Readme. 

We have that X.X.X reference already in the same file (see https://github.com/codecentric/chaos-monkey-spring-boot/blob/gh-pages/README.md#dependency-in-your-pom). So this PR brings consistency. 


We could change both X.X.X references to current version, but its easy to miss on each release (unless we have a release checklist).  I would preferr to do that (note is ok as well, but is polluting the readme unnecessary)

maiksensi:
Fine then. Keep it like it is. Let's merge this after the bugfix release.

WtfJoke:
Will create a new PR for chaning it to next release version (instead of X.X.X)

Issue #118, opened 2019-09-09
BeanCurrentlyInCreationException after upgrade to 2.1.0


smcgregor83:
After upgrading from 2.0.2 to 2.1.0 our Spring Boot applications will no longer start due to a BeanCurrentlyInCreationException.

## Expected Behavior
Spring Boot application should start.

## Current Behavior
Spring Boot application does not start.

## Possible Solution
Not sure. This only occurs when spring-cloud-starter-sleuth is also on the classpath.

## Steps to Reproduce
[demo.zip](https://github.com/codecentric/chaos-monkey-spring-boot/files/3589861/demo.zip)

## Context (Environment)
Automated upgrade of 3rd party library versions.

WtfJoke:
Hi Gregor

Thanks for your report. 👍 
I was able to reproduce it as well, using your demo.zip. The error doesnt occur with version 2.0.2. 
I am going to ask around in our team if somebody can have a look at it.

srempfer:
Narrowed down the problem to this library
```
        <dependency>
            <groupId>org.springframework.cloud</groupId>
            <artifactId>spring-cloud-context</artifactId>
        </dependency>
```
and the Auto-Configuration
```
org.springframework.cloud.autoconfigure.ConfigurationPropertiesRebinderAutoConfiguration.class
```

maiksensi:
We'll prepare a release in the next days.

WtfJoke:
Some analysis notes:

Tested by adding this dependency:
```
<dependency>
            <groupId>org.springframework.cloud</groupId>
            <artifactId>spring-cloud-context</artifactId>
            <version>2.1.3.RELEASE</version>
</dependency>
```

Issue was caused by  `org.springframework.cloud.context.properties.ConfigurationPropertiesBeans` and SpringComponentAspect (ChaosMonkeyComponentWatcher). 

In 2.0.2 watchers werent configurable at runtime and fixed with #84 (which kinda "reintroduced" this bug, but problem already existed earlier)

For more details see #120

WtfJoke:
Latest 2.2.0-SNAPSHOT should solve the issue, see https://codecentric.github.io/chaos-monkey-spring-boot/#snapshots

WtfJoke:
[2.2.1 released](https://github.com/codecentric/chaos-monkey-spring-boot/releases/tag/2.1.1) containing this bugfix

smcgregor83:
[...]

PR #121, opened 2019-10-01
Add context as dependecy to make sure cloud dependencies keep working

maiksensi:
**What**:
Add spring cloud dependency to make sure cloud dependencies keep working in the future.

**Why**:
see bug: #118 
Testing this directly seems too complicated and cumbersome. So, it makes sense to indirectly test the cloud dependencies for future compatibility. 

**How**:

**Checklist**:

- [x] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-spring-boot/pom.xml
+++ b/chaos-monkey-spring-boot/pom.xml
@@ -32,6 +32,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-sleuth</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-aspects</artifactId>
         </dependency>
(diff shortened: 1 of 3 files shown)


chaos-monkey-spring-boot/pom.xml
```diff
@@ -31,6 +31,11 @@
     </parent>
 
     <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-context</artifactId>
+            <scope>test</scope>
```
WtfJoke: Version is missing. Eg.  
```xml
<version>2.1.3.RELEASE</version>
```

maiksensi: I am not sure that we need the actual version here. If you look at all other dependencies - junit for example?

chaos-monkey-spring-boot/pom.xml
```diff
@@ -30,7 +30,25 @@
...
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
+        <dependency>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-starter-sleuth</artifactId>
+            <scope>test</scope>
```
WtfJoke: OCD triggered  :eyes: -->scope needs to be indented

Issue #140, opened 2019-10-19
Migrate tests to JUnit Jupiter / JUnit 5


Riggs333:
Migrate all tests in `chaos-monkey-spring-boot` to JUnit Jupiter.

- [ ] add JUnit Jupiter dependencies
- [ ] migrate tests to use annotations and api of `org.junit.jupiter`
- [ ] remove obsolete `public` modifier for test classes and test methods

Issue #90, opened 2019-08-13
Fix ChaosDemoApplicationTests#checkMetricsBean


WtfJoke:
The test (in demo project ext-jar) fails under windows and should be fixed

WtfJoke:
It fails also on linux when executed directly in IDE (intellij)

F43nd1r:
I don't think there is anything to fix here. This is by design. The ext-jar isn't available if it isn't added externally (e.g. in the IDE test run config).

I'm not sure if this can be tested in a way which doesn't require external configuration. I'd remove the test.

WtfJoke:
> I don't think there is anything to fix here. This is by design. The ext-jar isn't available if it isn't added externally (e.g. in the IDE test run config).
> 
> I'm not sure if this can be tested in a way which doesn't require external configuration. I'd remove the test.

Lets remove it 🔥

WtfJoke:
Fixed/removed with #262

PR #141, opened 2019-10-19
Migrate tests to JUnit Jupiter

Riggs333:
**What**:

First step to migrate all tests to JUnit Jupiter

Resolves: #140 

**Why**:

Stay up to date with latest JUnit and benefit from all the beauty. :tada: 

**How**:

Roughly followed the instructions in [JUnit documentation (Migrate from JUnit 4)](https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4).

**Checklist**:

- [ ] Documentation added N/A
      
- [x] Changelog updated
      
- [ ] Tests added N/A
      
- [x] Ready to be merged
      

I left the test class `ChaosMonkeyRequestScopeTest` out by now, but annotated a TODO.
This class fails with one test when migrating to the strict stubbing that comes with `MockitoExtension`.
(I suggest to use `@Nested` here, but that change would be too big for this PR.)

### DIFF
--- a/chaos-monkey-dependencies/pom.xml
+++ b/chaos-monkey-dependencies/pom.xml
@@ -96,15 +96,15 @@
                 <scope>provided</scope>
             </dependency>
             <dependency>
-                <groupId>org.junit.jupiter</groupId>
-                <artifactId>junit-jupiter-api</artifactId>
-                <version>${junit.jupiter.version}</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
                 <version>${junit.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-api</artifactId>
+                <version>${junit.jupiter.version}</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
--- a/chaos-monkey-dependencies/pom.xml
+++ b/chaos-monkey-dependencies/pom.xml
@@ -96,9 +96,27 @@
                 <scope>provided</scope>
             </dependency>
             <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-api</artifactId>
+                <version>${junit.jupiter.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
                 <version>${junit.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-engine</artifactId>
+                <version>${junit.jupiter.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.junit.vintage</groupId>
+                <artifactId>junit-vintage-engine</artifactId>
+                <version>${junit.vintage.version}</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
@@ -122,6 +140,12 @@
             <dependency>
                 <groupId>org.mockito</groupId>
                 <artifactId>mockito-core</artifactId>
+                <version>${mockito.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-junit-jupiter</artifactId>
                 <version>${mockito.version}</version>
                 <scope>test</scope>
             </dependency>
--- a/chaos-monkey-spring-boot/pom.xml
+++ b/chaos-monkey-spring-boot/pom.xml
@@ -63,14 +63,14 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <version>${junit.jupiter.version}</version>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit.jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/chaos-monkey-spring-boot/pom.xml
+++ b/chaos-monkey-spring-boot/pom.xml
@@ -63,8 +63,24 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit.jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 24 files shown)


chaos-monkey-dependencies/pom.xml
```diff
@@ -95,12 +95,30 @@
                 <version>${lombok.version}</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-api</artifactId>
+                <version>${junit.jupiter.version}</version>
+                <scope>test</scope>
+            </dependency>
             <dependency>
```
Riggs333: After all tests are migrated this dependency can be removed.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyRequestScopeSettingsTest.java
```diff
@@ -16,23 +16,23 @@
 
 package de.codecentric.spring.boot.chaos.monkey.configuration;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
-import static junit.framework.TestCase.assertTrue;
```
Riggs333: Replacing this import with `org.junit.Assert.assertTrue` is a suggestion by IntelliJ.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeySchedulerTest.java
```diff
@@ -86,8 +84,4 @@ public void shouldTriggerRuntimeScopeRunAttack() {
         return cronTask -> cronTask.getExpression().equals(schedule);
     }
 
-    @Before
-    public void initMocks() {
-        MockitoAnnotations.initMocks(this);
```
Riggs333: Since this class is extended with `MockitoExtension` the call is superfluous.
(I suspect it wasn't even need before.)

Riggs333:
Right now I have no idea why in the projec `chaos-monkey-demo-app-ext-jar` the test `ChaosDemoApplicationTests#checkMetricsBean` cannot find the `metrics` bean anymore. :thinking: 
Is the change of the maven-surefire-plugin version the root of this?

Riggs333:
Ok, looks like on the `master` branch with `maven-surefire-plugin` version `2.12.4` the build is not even executing the test class `ChaosDemoApplicationTests`. (Maybe because of the `Tests` suffix instead of `Test`?)

Any idea how to fix this?
I wonder why the test expects to find a bean with name `metrics` (of type `de.codecentric.spring.boot.chaos.monkey.component.Metrics`)?
Looking at the `ChaosMonkeyConfiguration` class this `Metrics` bean is only registered if `MeterRegistry` is on the classpath:
```java
    @Bean
    @ConditionalOnClass(name = "io.micrometer.core.instrument.MeterRegistry")
    public Metrics metrics() {
        return new Metrics();
    }
```

maiksensi:
Hey @Riggs333,
thanks for all the effort you put into this. It is much appreciated. However, I haven't had the chance to have a look at it. Actually, so far I never had a look at any of the metrics stuff. So, maybe @WtfJoke or @fletchgqc got some more experience. If not, I can't promise I get to it this week, maybe tomorrow or on friday.

WtfJoke:
Hi @Riggs333 again

Thanks for your work, appreciate the JUnit Upgrade 👍 
At first I thought, you probably just stumbled upon https://github.com/codecentric/chaos-monkey-spring-boot/issues/90, but after having looked at the build it fails on other subprojects as well.

If I execute `ChaosDemoApplicationTests` locally it works (on current master).  Maybe we need to replace 
```xml
        <dependency>
            <groupId>io.micrometer</groupId>
            <artifactId>micrometer-registry-prometheus</artifactId>
            <version>1.0.4</version>
        </dependency>
```
with micrometer core on subprojects to fulfill the condition.
I will also look into it during the next days.

WtfJoke:
> `master` branch with `maven-surefire-plugin` version `2.12.4` the build is not even executing the test class `ChaosDemoApplicationTests`. (Maybe because of the `Tests` suffix instead of `Test`?)

Can confirm that behavior. 
Your assumption about the suffix is correct (mvn test fails after changing it to `Test`). 

Updated #90 (since it fails on Linux as well).
My suggestion is to ignore this test(in demo-project ext-jar) in this PR and fix it with #90. Any other thoughts/ideas @maiksensi?




> Looking at the `ChaosMonkeyConfiguration` class this `Metrics` bean is only registered if `MeterRegistry` is on the classpath:

If we remove the condition, the test still fails. So I am not sure whats the cause of this.

chaos-monkey-dependencies/pom.xml
WtfJoke: Maybe I'm wrong but don't you mean the junit-vintage-engine?

chaos-monkey-spring-boot/pom.xml
```diff
@@ -62,11 +62,27 @@
             <artifactId>lombok</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit.jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
```
WtfJoke: Quite picky I know :see_no_evil: , but can you move that below of junit dependency (then its grouped with the other jupiter stuff)

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/assaults/ExceptionAssaultTest.java
```diff
@@ -20,96 +20,78 @@
...
+    MetricEventPublisher metricsMock;
 
     @Test
-    public void throwsRuntimeExceptionWithDefaultAssaultSettings() {
-        exception.expect(RuntimeException.class);
-
+    void throwsRuntimeExceptionWithDefaultAssaultSettings() {
         ExceptionAssault exceptionAssault = new ExceptionAssault(getChaosMonkeySettings(), metricsMock);
-        exceptionAssault.attack();
+        assertThrows(RuntimeException.class, exceptionAssault::attack);
```
WtfJoke: :+1:

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/assaults/MemoryAssaultIntegrationTest.java
```diff
@@ -22,7 +20,6 @@
 /**
  * @author Benjamin Wilms
  */
-@RunWith(SpringRunner.class)
```
WtfJoke: Glad it works without that. I thought it would need `@ExtendWith(SpringExtension.class)` or something similar

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScopeTest.java
```diff
@@ -35,6 +35,7 @@
 /**
  * @author Benjamin Wilms
  */
+// TODO migratie to @RunWith(MockitoExtension.class)
```
WtfJoke: `// TODO migrate to @ExtendWith(MockitoExtension.class)`

Typo in migrate and it should be ExtendWith instead of RunWith. Nothing serious, though :smiley:

chaos-monkey-dependencies/pom.xml
Riggs333: Yes, you're right, I mixed that up. :-)
This can even be cleaned up after moving to Spring Boot 2.2 (https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes#junit-5)

chaos-monkey-spring-boot/pom.xml
Riggs333: Sure, that's a good idea! :+1:

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScopeTest.java
Riggs333: Good catch! :bowing_man:

Riggs333:
Thanks for your feedback and reviews! :bowing_man: :+1: 
I pushed the fixes and also marked the failing test with `@Ignored` until issue #90 is solved.

PR #148, opened 2019-11-07
Created SpringRepositoryAspectJDBC for watching non-jpa repositories

ericwyles:
Created SpringRepositoryStereotypeAspect. Updated to auto configure and added and updated tests accordingly.

**What**:

**Why**:

**How**:

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [x] Tests added
      
- [ ] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -2,6 +2,7 @@
 == Changes in {project-version}
 
 === Bug Fixes
+- https://github.com/codecentric/chaos-monkey-spring-boot/issues/147[#147] Repository Watcher does not work on non-JPA Repositories
 
 === Improvements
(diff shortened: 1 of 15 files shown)


maiksensi:
@WtfJoke @ericwyles  can you please have a look at the naming of the aspects? 
@ericwyles would you mind adding an entry to the changelog 😄 please?

ericwyles:
@maiksensi The new naming looks great to me and I also updated the changelog. Please let me know if there is anything else I can do.

maiksensi:
> @maiksensi Or do we need an additional testcase? We had some case where all tests ran green without detecting a bug, right? Do you remember which case this was?

I don't think so. We need to have a look at that anyways, maybe document a bit better which app tests what (indirectly). In the meantime, I'll merge this. 
@ericwyles thanks for the patience to bear with the discussions, and  thank you for your contribution!

Issue #133, opened 2019-10-07
Use a autoformatting tool (integrate in pipeline too)


maiksensi:
In order to avoid "nitpicking" comments about the styling of a java file I suggest we integrate an autoformatting tool into our maven goals (which can be executed then) and also integrate that as a mandatory goal into our CI pipeline (if not correctly formatted fail pipeline).
In the best case, we only look at pure syntax formatting (like prettier does for JavaScript).

WtfJoke:
I suggest we should using [google-java-format](https://github.com/google/google-java-format) and using either this [revelc/formatter-maven-plugin](https://github.com/revelc/formatter-maven-plugin) (configurable maven format plugin) or this [coveooss/fmt-maven-plugin](https://github.com/coveooss/fmt-maven-plugin) (google style format maven plugin)

maiksensi:
As long as it works in the pipeline (checker) and in let's say intellij I really dont care as long as we get a consistent style by adding this.

WtfJoke:
https://github.com/diffplug/spotless looks neat and they also have a gradle/maven plugin. It supports google java format out of the box.

Seen in habitcentric (gradle):
```yaml
spotless {
    java {
        googleJavaFormat('1.7')
    }
}
```

fletchgqc:
I used google java format with fmt-maven-plugin on a recent project and it was a very positive experience. Although it takes a while to get used to the annoyance of your build being declared broken, it's fantastic to have a standard format which is enforced during the build chain. Diffs are much easier to read, etc.

I believe that google format is the right choice because it's good enough (actually very good) and by not thinking about it, we avoid having to have some kind of format selection process.

I have no experience with the other maven plugins. Happy to go with whatever others choose.

maiksensi:
I am happy to prepare a MR in the next days (already started tbh) including: 
- setup the plugin with checkstyle and a supression file (for disabled rules)
- setup running in the pipeline and failing it when violations occur
- docs explaining how to setup intellij 

Docs: 
https://google.github.io/styleguide/javaguide.html

They have an intellij plugin: 
https://raw.githubusercontent.com/google/styleguide/gh-pages/intellij-java-google-style.xml

IntelliJ Code Style File: 
https://raw.githubusercontent.com/google/styleguide/gh-pages/intellij-java-google-style.xml

WtfJoke:
Closed by #160

PR #160, opened 2020-01-02
Add spring boot formatter and checkstyle

maiksensi:
**What**:

Introduce google style guide as discused in #133 

**Why**:

see #133 - short: error prone formatting, avoid nitpicking

**How**:

- use checkstyle maven plugin and checkstyle dependency + configuration to enable google_checks

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [x] Tests added "N/A"
      
- [x] Ready to be merged


chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/ExceptionAssault.java
```diff
@@ -26,7 +26,7 @@
 /** @author Thorsten Deelmann */
 public class ExceptionAssault implements ChaosMonkeyRequestAssault {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(ExceptionAssault.class);
+  private static final Logger Logger = LoggerFactory.getLogger(ExceptionAssault.class);
```
fletchgqc: Actually we should use Lombok to get the logger, but we can do that on another PR.

pom.xml
```diff
@@ -15,61 +15,69 @@
...
-  <description>Chaos Monkey for Spring Boot parent project</description>
-
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
   <properties>
     <!-- project version -->
-    <revision>2.3.0-SNAPSHOT</revision>
+    <asciidoctor-maven-plugin.version>1.6.0</asciidoctor-maven-plugin.version>
```
WtfJoke: Was the revision deleted on purpose? And if so the comment would be wrong.

Else changes looks good and could be merged. 👍  I‘ll aprove it but question frim above should ve answered before we merge it.

maiksensi: The revision should only have been moved because of the formatting. Can you double check its there? I think it is now in: Line 61

maiksensi: @WtfJoke ping

maiksensi:
> Do we need to provide a link to a guide which shows how to install the code style?

Did you see: https://github.com/codecentric/chaos-monkey-spring-boot/pull/160/files#diff-bf6c05f66a41e4a8d614764c22eaf178R56 it includes a general explanation and also an explanation for intellij

pom.xml
WtfJoke: Yes youre right, its there. Sorry didnt have time earlier. 😞

WtfJoke:
> Did you see: https://github.com/codecentric/chaos-monkey-spring-boot/pull/160/files#diff-bf6c05f66a41e4a8d614764c22eaf178R56 it includes a general explanation and also an explanation for intellij

Yeah somehow I have seen it but ignored. Lets get it merged (as soon as latest changed are merged into)!

Sorry, that it took a while to review your last comments.

Issue #162, opened 2020-01-13
Change default values


fletchgqc:
CMSB defaults: 
chaos.monkey.enabled=false
chaos.monkey.watcher.service=true
chaos.monkey.assaults.latencyActive=true

This is pretty unintuitive. Because:
1) Why would you want latency and service on by default? Everything off by default would be more logical.
2) What about enabling chaos monkey by default? You put the profile there, you obviously want CMSB. If the attacks are all disabled by default, then you could enable CMSB by default. This reduces the barriers to entry, because it's kinda confusing that you have to activate the profile AND enable it AND configure the attacks you want. Too many sources of error for newcomer. The only thing that stands against it is a possible runtime cost, when people want the profile running on all their apps by default (?).

maiksensi:
I agree, the default values do not make sense, your proposal looks good to me. 
What I like is that you have to enable the profile and then chaos monkey itself. This is handy in case I have a Gameday, we enable the profile to configure chaos monkey via rest and then we enable it at a specific time. So, I would argue that the workflow, activating the profile, configuring the attacks, enabling chaos monkey makes sense. 
Regarding the bigger picture:
- I dont like that some attacks are cron based, all attacks should be able to run via cron, but also once in case cron is not configured. 
- I dont like that we do not have a simple get started tutorial which explains that you first activate the monkey with basic default (everything deactivated) settings, then you configure the monkey and its attacks and then you activate it (cron expression would be a optional configuration in that case).
Regarding the issue, changing the default values, that makes sense imho, should we also tackle a get started tutorial?

fletchgqc:
on/off switch: OK

cron: I have some thoughts on that. In any case, this is a separate issue, let's handle it somewhere else.

Tutorial: Manuel and I basically made one in audio/visual form, we are giving it at conferences. It kind of conveniently avoids the default issue. We could either use it as-is from, e.g. conf42 (video will be posted circa March 2020), or we could just record it again specially for us. Recording it again would not be _too_ much work since we already have the material. Or do you mean rather written?

PR #164, opened 2020-01-28
Change default configuration to everything off and predictable values

fletchgqc:
Change default configuration to everything off and predictable values - https://github.com/codecentric/chaos-monkey-spring-boot/issues/162

**Checklist**:

- [x ] Documentation added
- [x ] Changelog updated
- [x ] Tests added
- [x ] Ready to be merged

### DIFF
--- a/chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/DefaultSettingsIntegrationTest.java
+++ b/chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/DefaultSettingsIntegrationTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/DefaultSettingsIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2018 the original author or authors.
```
maiksensi: Still not 100 % sure whether this should be 2020

WtfJoke: Its correct. I read in the internet that its the starting year the copyright is valid. If you change it to 2020 its not copyrighted in 2019 for example.

Source: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

fletchgqc:
I would argue that we should remove the copyright statement altogether. As mentioned in the comments on that SO answer, it is automatic and has no bearing for us.

PR #166, opened 2020-01-30
Describe the API docs with OpenAPI, validate OpenAPI during build

fletchgqc:
**What**:
We once talked about describing the API docs with OpenAPI. I have started this by documenting the watcher GET request, and including Swagger UI to display the documentation nicely. Also, the OpenAPI documentation is validated during the build using the Atlassian Swagger Request Validator library. This allowed me to add OpenAPI validation to our existing functional tests. The asciidoc links to the Swagger UI page.

**Why**:
This is debatable, but I think it's better to write your API doco by hand than to generate it. However, writing by hand is error-prone, this is where automatic validation is necessary to solve this.

I have not documented the entire API yet. I wanted to get the full chain in place for one endpoint first. This means: 
- OpenAPI file written.
- Swagger UI running on the gitlab pages site and correctly showing the OpenAPI file.
- Verification of the OpenAPI in the relevant functional test (just one test right now, because just one endpoint is documented).
- Ascii doc changed to link to OpenAPI.

Although a bit strange to only have one endpoint in OpenAPI, this is self-contained and mergeable as-is. I would like to see the maintainers agree that this is correctly done and merge the PR. Then we could go ahead with adding the rest of the endpoints as follows for each endpoint:
1. add the endpoint doco to OpenAPI file.
1. modify the relevant functional test to include OpenAPI verification.
1. remove this endpoint doco from the asciidoc pages and link to swagger-ui instead.

By the way, I also want to clean up and update the functional tests a little. Another reason why I want to get this PR finished first, then I can attach that separately.

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [x] Tests added
      
- [x] Ready to be merged


chaos-monkey-docs/src/main/swagger-ui/oauth2-redirect.html
```diff
@@ -0,0 +1,68 @@
+<!doctype html>
```
WtfJoke: Is this file generated?

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/endpoints/ChaosMonkeyRequestScopeRestEndpointIntTest.java
```diff
@@ -99,11 +114,16 @@ void postChaosMonkeySettingsValueObjectWatcherPropertiesNULL() {
...
+        // openapi validation
+        testRestTemplate.getRestTemplate().setInterceptors(Collections.singletonList(validationInterceptor));
+
         ResponseEntity<WatcherProperties> result =
                 testRestTemplate.getForEntity(baseUrl + "/watchers", WatcherProperties.class);
 
         assertEquals(HttpStatus.OK, result.getStatusCode());
         assertEquals(chaosMonkeySettings.getWatcherProperties().toString(), result.getBody().toString());
+
+        testRestTemplate.getRestTemplate().setInterceptors(Collections.emptyList());
```
WtfJoke: I think this should be better done in tearDown. In case of an exception this wont be called/rolled back. The same could apply for setting the interceptors (in case its wanted for each test)

WtfJoke:
I havent had any strong opinion about having the documentation generated or a semi-automatic approach like this (handwritten and automated validation). 

The good thing with both ways its consistent documented in there and not half in regular documenation and other half somewhere else. I like the approach of having one documentation and add the "since v2.4.0" approach. You will loose then, however to make document fundemental api changes.

chaos-monkey-docs/src/main/swagger-ui/oauth2-redirect.html
fletchgqc: Everything in this commit comes from Swagger-UI, i.e. https://github.com/swagger-api/swagger-ui/tree/master/dist. I separated the commits so that it's easier to see what I wrote and what's just copied in.

Swagger-UI is a tool which presents OpenAPI files in an appealing format (i.e., in a webpage).

fletchgqc:
> I like the approach of having one documentation and add the "since v2.4.0" approach.

This particular point is unrelated to this PR. Let's have that discussion elsewhere, please.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/endpoints/ChaosMonkeyRequestScopeRestEndpointIntTest.java
fletchgqc: Good point, removing interceptors added in df10163b99e532451b418a9e20ef30e46d588a7d.

Yes, adding the interceptors will be re-evaluated when I work over the whole file and move more things to OpenAPI (next PR, after you approve this one :-)).

chaos-monkey-docs/src/main/swagger-ui/oauth2-redirect.html
WtfJoke: Yeah I know swagger, but I wasnt sure about the oauth2 stuff :D

fletchgqc:
Let’s hold on this for a moment. I’m looking into another option which is called rest docs. It might be preferable. I’ll update this PR and present the two options to you for comment.

PR #175, opened 2020-04-17
Introduce proper formatter

maiksensi:
Signed-off-by: Maik Figura <maik.figura@codecentric.de>

**What**:
After #160 horribly failed to introduce more ease of development we decided to look for another solution that would work like prettier (javascript) or black (python) 

**How**:
- introduce spotless as a maven goal and as a step in ci pipeline

**Checklist**:

- [x] Documentation added
- [x] Changelog updated
- [ ] Tests added - Not needed.
- [x] Ready to be merged


.travis.yml
```diff
@@ -9,7 +9,9 @@ cache:
   directories:
     - $HOME/.m2
 install: ./mvnw install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-script: ./mvnw verify -B
+script:
+  - ./mvnw spotless:apply
```
WtfJoke, line 13: I think you wanted to run `mvn spotless:check` instead of `apply`. Apply doesnt make much sense in CI as long we dont commit the changes spotless applied.

As this is already run with mvnw verify, the change is probably not needed.

CONTRIBUTING.adoc
```diff
@@ -53,21 +53,18 @@ To build and test the code you can run
...
+run via `./mvnw spotless:check`. Use `./mvnw spotless:apply` to fix these violations.
 
 [NOTE]
 ====
-`./mvnw verify` also runs checkstyle while `./mvnw test` does not.
+`./mvnw verify` also runs spotless while `./mvnw test` does not.
 ====
 
 === IntelliJ Setup
-If you're using IntelliJ then we highly recommend to install the https://plugins.jetbrains.com/plugin/8527-google-java-format[google format plugin] via the plugin marketplace.
```
WtfJoke, line 69: 👿 ❤️  (didnt like the plugin)

pom.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!--
-  ~ Copyright 2014-2018 the original author or authors.
+  ~ Copyright 2020 the original author or authors.
```
WtfJoke, line 3: Shouldn't we stick with original dates, eg. 2014? I red here https://stackoverflow.com/a/2391555/2970422 that you give away your copyright for previous years like that. 

We can go for 2020 and later on adjust it again to 2014? When we add the licence header in pom.xml

.travis.yml
maiksensi: yep! good catch!

CONTRIBUTING.adoc
maiksensi: yah that's why I removed that section and added that one should install the style guide file.

pom.xml
maiksensi: I will revert those changes. I am voting for removing them completely, you can see the history via GIT anyways? (out of scope for this PR I guess)

maiksensi: On a side note: 2014 seems a bit early :)

PR #180, opened 2020-05-15
Fix #72 - Serializing issue with spring cloud dependencies

### LINKED ISSUE #72: I have an error using this path("/actuator/chaosmonkey");Is there a serialization problem?
**{
    "timestamp": "2019-06-03T12:31:34.711+0000",
    "status": 500,
    "error": "Internal Server Error",
    "message": "Type definition error: [simple type, class org.springframework.validation.beanvalidation.MethodValidationInterceptor]; nested exception is com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class org.springframework.validation.beanvalidation.MethodValidationInterceptor and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: de.codecentric.spring.boot.chaos.monkey.configuration.ChaosMonkeySettings[\"assaultProperties\"]->de.codecentric.spring.boot.chaos.monkey.configuration.AssaultProperties$$EnhancerBySpringCGLIB$$6141c1b[\"advisors\"]->org.springframework.aop.support.DefaultPointcutAdvisor[0]->org.springframework.aop.support.DefaultPointcutAdvisor[\"advice\"])",
    "path": "/actuator/chaosmonkey"
}**

WtfJoke:
**What**:
Fixes #72 by introducing a `ChaosMonkeySettingsDto` which returns an `AssaultPropertiesUpdate` (which is kinda a DTO too) instead of our existing internal config object `AssaultProperties`.

**Why**:
If somebody has Spring Cloud Dependencies and use the actuator endpoint the endpoint can have problems serializing the AssaultProperties Object.  Spring @ConfigurationProperties  beans will be proxied so that the underlying can be rebound/recreated when a configuration refresh is issued. This causes that the deserialized objects has additional properties (from the proxy, see screenshot) or cannot even be serialized like described in #72

**How**:
By not returning our ConfigurationProperties directly and abstract it in an Dto (AssaultPropertiesUpdate)

**Notes**:
On this screenshot you can see the properties which get added by the spring proxy and accidently returned
grafik
I dont know if we can write test for this change without huge efforts (since it relies, that on a non existing spring cloud dependency). 

I verified the behaviour with an [sample app](https://github.com/fletchgqc/mediator/tree/cmsb%2372serialization) with `2.2.0` I get above result. With `2.3.0-SNAPSHOT` (this pr locally build) I get the correct result. 

**Checklist**:

- [N/A] Documentation added
      
- [x] Changelog updated
      
- [N/A] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -3,6 +3,7 @@
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
+- https://github.com/codecentric/chaos-monkey-spring-boot/pull/180[#180] Fixes serialization issues when using actuator combined with spring cloud dependencies (see https://github.com/codecentric/chaos-monkey-spring-boot/issues/72[#72] for more details)
 
 === Improvements
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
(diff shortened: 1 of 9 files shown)


fletchgqc:
Using DTOs is generally a good idea. So approval from me.

Issue #181, opened 2020-06-03
Feature request: random factor for runtime assault scheduling


mbouchenoire:
Hello,

I see two issues with the current runtime assault scheduling:

1. It's predictable since we know when the assaults will take place
2. Using Kubernetes when pods are started - almost- simultaneously (which happen the majority of the time) the assaults also happen - almost - simultaneously, which is not really what will happen without this library (even though it could, yes).

Edit : I know that it's possible to use SpEL to randomize the cron at startup, but it's still fixed when the application is started :(

Your thoughts on this?

maiksensi:
Hey @mbouchenoire,
the described behaviour is partly intended. When this was implemented we wanted it to be predictable so correlating logs and traces would be easy. I see your point that there are use cases where you don't want it that predictable, but I guess you can help yourself with SpEL. 
For the second part, I normally keep CMSB deactivated when I start up my applications. When everything is settled and the app reaches steady state, I activate CMSB via `/enable`. @fletchgqc @WtfJoke what do you folks think? Is the immediate run of the assault a bug? I would say it is intended depending on your settings. Isn't the linux cron behaving the same way? 

But I also see your point @mbouchenoire if the app is pre-configured this is not very handy.

WtfJoke:
Hi @mbouchenoire 

1. I agree with @maiksensi on first point, that the runtime assaults were made to be predictable (either using cron expressions or calling the `/assaults/runtime/attack` rest endpoint).

2. Usually I use CMSB also with the `/enable` endpoint (like maik described). 
The behaviour you describe shouldn't apply to Runtimeassaults (killApp/memory) in my opinion. 
As far as I am aware of these runtime assaults only trigger like described in 1. (therefore usually not on container/pod start). 

Can you share your application.properties/yaml (e.g CSM config)? 
Did you maybe confuse it with request assaults (latency, exception)? These could trigger on application start depending on your environment/config.

mbouchenoire:
I understand that the current behavior is intented, but I'm advocating for an alternative, being *actual* chaos :).

On an other note, I don't know if it's related to this issue but I think it is : I'd to enable the KillApp and Memory assaults with different timings (even if not random). Currently, as far as I understand it you cannot schedule those 2 assaults independently, thus you cannot use them simultaneouslywith a cron (even with a random one).

fletchgqc:
[...]

PR #183, opened 2020-06-17
Add dedicated cron for AppKiller and Memory assaults

mbouchenoire:
**What**:

Basically, this PR adds 2 configuration keys:
- `chaos.monkey.assaults.killApplication.cron.expression`
- `chaos.monkey.assaults.memory.cron.expression`

They act the same way as the currently existing `chaos.monkey.assaults.runtime.scope.assault.cron.expression`. If a "specialized" cron expression is specified, it will override the "global" cron expression.

**Why**:

This PR is a followup to the discussion that happened here: https://github.com/codecentric/chaos-monkey-spring-boot/issues/181

As @maiksensi pointed out, it would be something nice to at least discuss. I figured a pull request would be a more concrete starting point :). We can discuss this in a issue if necessary.

**How**:

The runtime assaults are now scheduled separately, and it's pretty much it !

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [x] Tests added
      
- [x] Ready to be merged


maiksensi:
Hey @mbouchenoire,
thanks for the contribution ❤️. I already had a fast look at it and I think the overall state is quite good 👍 🚀 . Give me some time to have a better look at it in the next days. I think I also wanna try it out locally first before merging. Did you try your changes with your use case already?

maiksensi:
Okay, so I had another go at it. Let's get to it: 

- How can I post a random cron via REST? 
--> POST to `{{site}}/actuator/chaosmonkey/assaults` with payload: `{
	"killApplicationCronExpression": "0 ${random.int[0,30]} 4 * * ?"
}`
fails with `"message": "For input string: \"${random.int[0\"",`.

I also tested both settings a bit with the REST api and the demo app (https://github.com/codecentric/chaos-monkey-spring-boot/tree/master/demo-apps/chaos-monkey-demo-app) incrementally enabling `memoryAssault` and then `killAppAssault`. That seems to break stuff. 
Let's say you enable `memoryAssault` with a POST to `{{site}}/actuator/chaosmonkey/assaults`: 

```
{
"memoryCronExpression": "*/3 * * * * ?",
"memoryActive": true
}
```

then wait a bit, and then enable `killAppAssault`:

```
{
"killApplicationActive": true,
"killApplicationCronExpression": "*/1 * * * * ?"
}
```

then neither the memory is filled nor is the app killed. Are they executed sequentially? Maybe the memory assault is not finishing. 

On a side note, I think we should remove `"runtimeAssaultCronExpression": "OFF"` completely in favour of the two other settings. I don't see any value for users in that settings. It only complicates things.

mbouchenoire:
> How can I post a random cron via REST?

This is not made possible with this PR. The random capabilities are identical to the current ones with `runtimeAssaultCronExpression`, which is only possible using SpEL with with startup configuration.

> then neither the memory is filled nor is the app killed. Are they executed sequentially? Maybe the memory assault is not finishing.

There is indeed something to fix here, scheduling a new assault will un-schedule the previous ones. I'll work on this.

> On a side note, I think we should remove "runtimeAssaultCronExpression": "OFF" completely in favour of the two other settings. I don't see any value for users in that settings. It only complicates things.

I agree, maybe we should keep it for backward compatibility and remove it in the next major version ?

WtfJoke:
Hey @mbouchenoire 

We've just gone through the open pull requests. 

> There is indeed something to fix here, scheduling a new assault will un-schedule the previous ones. I'll work on this.

Are you still interested to work on this?

> I agree, maybe we should keep it for backward compatibility and remove it in the next major version ?

Yeah we could deprecate it now and remove it on the next major/minor version.

mbouchenoire:
> Are you still interested to work on this?

Sorry for the delay, I've been quite busy since then :crying_cat_face:.
Unfortunately, I cannot guarantee to work on this until at least next year, but I'd be glad to if the work is still up.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/AssaultProperties.java
```diff
@@ -56,14 +60,15 @@
...
+  private String memoryCronExpression;
+
   private volatile boolean cpuActive = false;
 
   private int cpuMillisecondsHoldLoad = 90000;
 
   private double cpuLoadTargetFraction = 0.9;
 
-  @Value("${chaos.monkey.assaults.runtime.scope.assault.cron.expression:OFF}")
-  private String runtimeAssaultCronExpression;
```
WtfJoke, line 66: I think we should keep the property and make it deprecated and give people time to stop using it and eventually remove it in the next version.

Until we deprecate it we can use the individual expressions to override the runtimeAssaultCronExpressions. Eg. if you define a global one, you can still override it on cpuCronExpression level. 

But there most likely also easier alternatives

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/AssaultProperties.java
```diff
@@ -46,6 +48,8 @@
 
   private boolean killApplicationActive = false;
 
+  private String killApplicationCronExpression;
```
WtfJoke, line 51: The new cronExpressions are missing a default value. Eg. "OFF"?

I saw that we don't consider null values, but it should align with the docs. Also the user should know how he can disable an expression (so I guess having the default value "OFF" is the better choice).

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyScheduler.java
```diff
@@ -1,49 +1,83 @@
...
     }
-
-    if (active) {
-      CronTask task = new CronTask(runtimeScope::callChaosMonkey, cronExpression);
-      currentTask = scheduler.scheduleCronTask(task);
+    if (!currentTasks.isEmpty()) {
+      for (Iterator<Map.Entry<ChaosMonkeyRuntimeAssault, String>> iterator =
+              cronExpressions.entrySet().iterator();
+          iterator.hasNext(); ) {
+        Map.Entry<ChaosMonkeyRuntimeAssault, String> entry = iterator.next();
```
WtfJoke, line 47: Maybe here is a good way to use [val](https://projectlombok.org/features/val) from lombok.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyScheduler.java
```diff
@@ -1,49 +1,83 @@
...
+        ScheduledTask task = currentTasks.get(entry.getKey());
+        if (task != null) {
+          if (Objects.equals(((CronTask) task.getTask()).getExpression(), entry.getValue())) {
```
WtfJoke, line 50: Dont like this wild cast here (allthough we should be safe here as we only put CronTasks :D), but not really sure if we can improve it.
When we would do an instance check it will be typesafe but might lead to bugs.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyScheduler.java
```diff
@@ -1,49 +1,83 @@
...
 
   public void reloadConfig() {
-    String cronExpression = config.getRuntimeAssaultCronExpression();
-    boolean active = !"OFF".equals(cronExpression);
-
-    if (currentTask != null) {
-      Logger.info("Cancelling previous task");
-      currentTask.cancel();
-      currentTask = null;
+    Map<ChaosMonkeyRuntimeAssault, String> cronExpressions = new LinkedHashMap<>();
```
WtfJoke, line 39: Its hard to follow the code in the method. Splitting the code into multiple methods might help. 

1. load expressions
2. filter/removing expressions which are already running/scheduled
3. cancel tasks which needs to be rescheduled
4. reschedule tasks

Currently 2 and 3 is combined, it might be not the best thing (maybe it is, but I would like to compare it how it looks)

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/AssaultPropertiesUpdate.java
```diff
@@ -83,7 +87,7 @@
   @DecimalMin("0.1")
   private Double cpuLoadTargetFraction;
 
-  @Nullable private String runtimeAssaultCronExpression;
+  @Nullable private String cpuCronExpression;
```
WtfJoke, line 90: When we avoid having null values (see previous comment), we can remove the `@Nullable` annotation

WtfJoke (changes requested):
Thanks for bringing this up to date @F43nd1r. 🙏 

I've made some remarks. As stated before I think we should keep the existing global assault cron expression and deprecate it, before we remove it. 

Also I think that the reloadConfig method should receive some more love in order to make it easier readable.

In total it looks promisable!  👍

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/AssaultPropertiesUpdate.java
F43nd1r: `@Nullable` on `AssaultPropertiesUpdate` should always stay, as a null value here means "don't change"

WtfJoke: true

WtfJoke (approved):
LGTM 🪂

Issue #184, opened 2020-08-03
ChaosMonkeyConfiguration class issue while injecting Parameter 0 of method scheduler with consul Spring Boot


smreddy2004:
I have a Spring Boot Hoxton.SR6 application and when I start to use both chaos-money and consul-discovery

		<dependency>
			<groupId>de.codecentric</groupId>
			<artifactId>chaos-monkey-spring-boot</artifactId>
			<version>2.2.0</version>
		</dependency>

		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-consul-discovery</artifactId>
		</dependency>

## Expected Behavior
It should ideally pick the scheduler even if more than one such instance exists. 

## Current Behavior
It fails with and exception 
Description:

Parameter 0 of method scheduler in de.codecentric.spring.boot.chaos.monkey.configuration.ChaosMonkeyConfiguration required a single bean, but 2 were found:
	- taskScheduler: defined by method 'taskScheduler' in class path resource [org/springframework/boot/autoconfigure/task/TaskSchedulingAutoConfiguration.class]
	- catalogWatchTaskScheduler: defined by method 'catalogWatchTaskScheduler' in class path resource [org/springframework/cloud/consul/discovery/ConsulCatalogWatchAutoConfiguration.class]

## Possible Solution
It is unable to choose between the 2 schedulers available. Looking at the scheduler method of ChaosMonkeyConfiguration, may be it can be solved by using Qualifier to inject specific scheduler like "taskScheduler"

## Steps to Reproduce

[SampleProjectToReproduceError.zip](https://github.com/codecentric/chaos-monkey-spring-boot/files/5014604/SampleProjectToReproduceError.zip)

Import the zip as Maven project and start running as Spring Boot application.

maiksensi:
Hey @smreddy2004, 

I didn't have the time to look at the problem in detail, but if you have an idea on how to fix it, feel free to open a MR and we can discuss details there. Do you think your proposed fix will have side effects?

smreddy2004:
Hi @maiksensi ,

I don't think the fix will have any side effects. Today by default it autowires 'taskScheduler' bean since it is the only TaskScheduler bean available. But with consul in the classpath now there is conflict because of another TaskScheduler bean available named 'catalogWatchTaskScheduler'.

[...]

PR #186, opened 2020-08-13
Use chaos monkey specific scheduler to avoid multiple scheduler confl…

smreddy2004:
Use chaos monkey specific scheduler to avoid multiple scheduler conflict issue when we have more than one scheduler already defined in the application code

**What**: Created a new TaskScheduler named chaosMonkeyScheduler.

**Why**: When we have more than one scheduler already defined in the application code like say "taskScheduler" created by spring boot and "catalogWatchTaskScheduler" by consul, the chaosMonkey to autowire the scheduler it fails with exception "ChaosMonkeyConfiguration required a single bean, but 2 were found:".

This also gives us facility to tune our scheduler accordingly if needed in future.

**How**: Added a new Bean named "chaosMonkeyTaskScheduler" and using it to create ChaosMonkeyScheduler in scheduler(...) method with Qualifier attribute as chaosMonkeyTaskScheduler on TaskScheduler parameter.

**Checklist**:

- [ ] Documentation added N/A
      
- [ x] Changelog updated
      
- [ ] Tests added N/A
      
- [ x] Ready to be merged

### DIFF
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
@@ -144,7 +144,6 @@ public class ChaosMonkeyConfiguration {
   public ChaosMonkeyScheduler scheduler(
       @Nullable @Qualifier(CHAOS_MONKEY_TASK_SCHEDULER) TaskScheduler scheduler,
       ChaosMonkeyRuntimeScope runtimeScope) {
-    Logger.info("Creating chaos monkey scheduler with scheduler  : " + scheduler);
     ScheduledTaskRegistrar registrar = null;
     if (scheduler != null) {
       registrar = new ScheduledTaskRegistrar();
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
@@ -52,7 +52,6 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
 import org.springframework.context.annotation.Profile;
 import org.springframework.core.io.ClassPathResource;
-import org.springframework.lang.Nullable;
 import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.annotation.EnableScheduling;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
@@ -142,13 +141,10 @@ public class ChaosMonkeyConfiguration {
 
   @Bean
   public ChaosMonkeyScheduler scheduler(
-      @Nullable @Qualifier(CHAOS_MONKEY_TASK_SCHEDULER) TaskScheduler scheduler,
+      @Qualifier(CHAOS_MONKEY_TASK_SCHEDULER) TaskScheduler scheduler,
       ChaosMonkeyRuntimeScope runtimeScope) {
-    ScheduledTaskRegistrar registrar = null;
-    if (scheduler != null) {
-      registrar = new ScheduledTaskRegistrar();
-      registrar.setTaskScheduler(scheduler);
-    }
+    ScheduledTaskRegistrar registrar = new ScheduledTaskRegistrar();
+    registrar.setTaskScheduler(scheduler);
     return new ChaosMonkeyScheduler(registrar, assaultProperties, runtimeScope);
   }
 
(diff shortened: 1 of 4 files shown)


maiksensi:
Hey @smreddy2004 that looks good to me! Thank you for the contribution. I have one question. Would we use `@Qualifier` here or `@Primary`? @WtfJoke can you have a short look at it too please? Besides that, as soon as it is merged, I think I can make a release at the end of the week.

smreddy2004:
@maiksensi I believe `@Qualifer` is a better option. One reason, if the client application already has declared a TaskScheduler and marked it as `Primary`, we would again run into similar issue and now the exception would be `Caused by: org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'org.springframework.scheduling.TaskScheduler' available:  more than one 'primary' bean found among candidates`. 

And we don't want client to forcefully remove their `@Primary` annotation for us so things work and it might not even in their control if it annotation is in a notherthird party lib they are using.

maiksensi:
okay. So, if I understand this correctly: we're always using our `ChaosMonkeySheduler` in case multiple `TaskSheduler`s are present?

smreddy2004:
@maiksensi We will be using `chaosMonkeyScheduler` in all the cases. We won't rely on a scheduler to be provided by application. Because the issue is, today if the application has more than one scheduler configured then we won't know which one to use and are failing with an exception `chaosMonkey to autowire the scheduler it fails with exception "ChaosMonkeyConfiguration required a single bean, but 2 were found`. 

In my test case, "taskScheduler" was created by spring boot and "catalogWatchTaskScheduler" was by consul, and we don't know which ones to use and thus the above error.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
```diff
@@ -138,7 +142,8 @@ public ChaosMonkeyRequestScope chaosMonkeyRequestScope(
 
   @Bean
   public ChaosMonkeyScheduler scheduler(
-      @Nullable TaskScheduler scheduler, ChaosMonkeyRuntimeScope runtimeScope) {
+      @Nullable @Qualifier(CHAOS_MONKEY_TASK_SCHEDULER) TaskScheduler scheduler,
```
WtfJoke, line 145: Probably `@Nullable` is not necessary any more since we inject the bean always in method `chaosMonkeyTaskScheduler`.

Therefore I would suggest to remove it

```suggestion
      @Qualifier(CHAOS_MONKEY_TASK_SCHEDULER) TaskScheduler scheduler,
```

WtfJoke:
Hi @smreddy2004 
Thank you for your contributon. 

Looks quite straight forward.
This will fix issues like #184 and similar ones. 
I think this is fine like that (despite my little question/code suggestion). 

Probably a lot of logic/null-checks in `ChaosMonkeyScheduler` are probably not needed anymore, since we always have a Scheduler now. 
Should we get rid of this code in this PR as well? What do you think @smreddy2004 @maiksensi?

I talk mainly about these two code-blocks:
```java
    if (scheduler == null) {
      Logger.warn(
          "No ScheduledTaskRegistrar available in application context, scheduler is not functional");
    }
```

```java
   if (scheduler == null) {
        // We might consider an exception here, since the user intent could
        // clearly not be serviced
        Logger.error("No scheduler available in application context, will not process schedule");
```

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
smreddy2004: yeah Nullable is not needed now, will remove it.

smreddy2004:
@WtfJoke I think we should remove the logic related to null-checks as a part of this PR. Not point having the dead code.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
smreddy2004: Removed nullable.

smreddy2004:
@WtfJoke @maiksensi I have removed the null checks and related unit test. You can have a look at https://github.com/codecentric/chaos-monkey-spring-boot/pull/186/commits/b5baac02fce503a9b7aaada854baa6da56a3dc7a

maiksensi:
ahhh, it was good to ask @WtfJoke to also have a look. I wasn't sure about the Null-Checks. Well spotted! ❤️  @smreddy2004 that looks great. Nice that you also removed the corresponding test which is not needed anymore. I guess this can be merged. 👍

WtfJoke (approved):
LGTM

smreddy2004:
@WtfJoke I would think, it is follow the design principle of inject the dependencies rather than creating the dependencies on own.

maiksensi:
Yeah I agree with @smreddy2004. But I wanted to congratulate 👏  you on your first contribution (at least that is what Github is telling me :)). Thank you! 👍

smreddy2004:
Thanks @maiksensi 👍

PR #191, opened 2020-11-06
Problem: Proxying final component classes results in IllegalArgumentException

### LINKED ISSUE #188: Enhancement to automatically ignore Final classes in watcher Aspect classes
Hi,
Currently I am using spring boot version 2.1.6 and chaos monkey version 2.1.1

I have a final class with @Component annotation, unfortunately it is imported from a third party library and thus cannot be made "non-final".

**The issue I am facing is as follows:**
When I start my application the **SpringComponentAspect** tries to do a pointcut on all the classes in my application that have a @Component annotation, but since one of my component classes is final, it results in an exception.
"Could not generate CGLIB subclass of class com.test.abc: Common causes of this problem include using a final class or a non-visible class; nested exception is java.lang.IllegalArgumentException: Cannot subclass final class com.test.abc"

I understand that this exception is not from chaos monkey but thrown from Spring AOP. However, this puts me in a deadlock as I can't change the third party code that I've imported.

I was wondering if the Aspect classes in de.codecentric.spring.boot.chaos.monkey.watcher package be refactored so as to ignore final classes?

**Resolution**:  @Pointcut(within(@org.springframework.stereotype.Component *) && !within(is(FinalType))

Also the same can be applied for other pointcuts as well e.g: controller aspect, service aspect, etc.

Any help on this is appreciated.

sappo:
Solution: Exclude final component classes from being proxied

**What**:
Fixes #188

**Why**:
Choas monkey's watcher aspect will try to proxy every spring component. This attempt will fail for final classes though.

**How**:
Use an pointcut expression to exclude all final spring components

**Checklist**:

- [x] Documentation added - Not necessary
      
- [x] Changelog updated
      
- [x] Tests added
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/SpringComponentAspectIntegrationTest.java
+++ b/chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/SpringComponentAspectIntegrationTest.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package de.codecentric.spring.boot.chaos.monkey.watcher;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import de.codecentric.spring.boot.chaos.monkey.component.ChaosMonkeyRequestScope;
+import de.codecentric.spring.boot.chaos.monkey.component.MetricEventPublisher;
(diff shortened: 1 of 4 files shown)


chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/SpringComponentAspectIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
+
+/**
+ * @author Kevin Sapper
+ */
+@SpringBootTest
+class SpringComponentAspectIntegrationTest {
```
fletchgqc, line 40: I find the the amount of tests does not correlate with the change.

I mean: it's a global change; you can test that with one test, or if you really want, then two to make sure it applies more globally. Consider that every test requires maintenance later, so it's better to have as few tests as possible to test everything which you want to. What do you think?

fletchgqc (changes requested):
Great stuff.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/SpringComponentAspectIntegrationTest.java
sappo: Sure, in the end everything is a component in spring. I'll drop the tests for service, repository and controller and restcontroller

fletchgqc (approved):

PR #210, opened 2021-03-24
Problem: Environment properties can clash with chaos monkey properties and prevent app start up

### LINKED ISSUE #209: ENV properties can clash with chaos monkey properties and prevent app from starting up
## Expected Behavior

chaos-monkey-spring-boot properties should not clash with any other client app properties

## Current Behavior

Environment properties currently have a name clash with chaos monkey properties (i.e. **ENABLED=foo** clashes with **chaos.monkey.enabled**).  Here is the top level error if the client app is run with **`ENABLED=foo`** environment variable:
```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'de.codecentric.spring.boot.chaos.monkey.configuration.ChaosMonkeyConfiguration': Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'chaos.monkey-de.codecentric.spring.boot.chaos.monkey.configuration.ChaosMonkeyProperties': Unsatisfied dependency expressed through field 'enabled'; nested exception is org.springframework.beans.TypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'boolean'; nested exception is java.lang.IllegalArgumentException: Invalid boolean value [foo]
```

## Steps to Reproduce

run any of the demo apps in this repo with clashing environment variables.
For example:
```ENABLED=foo;COMPONENT=bar;SERVICE=baz...```

## Context (Environment)

Environment variables are used heavily within many of the tools used to build and deploy cloud hosted applications. 
Currently there is significant potential for property name clashes and this should be eliminated.

### PR DESCRIPTION
**Pull request by alexfincham**
**What**:
Fixes #209 

**Why**:
Currently when a client app is run with Environment variables that have the same name as any chaos monkey properties, the ENV variables take precedence and clash with the chaos monkey properties due to the SpEL expressions used with **`@Value`** annotation.  This can prevent the client app from starting up.

**How**:
**`@ConfigurationProperties`** uses setter injection for config properties by default so the use of field injection through the **`@Value`** annotation within classes already annotated by **`@ConfigurationProperties`** is redundant and also error prone.  The problem was solved through removing all of these redundant uses of **`@Value`** and using the full property path for the one case where property path does not directly map to property setter (field **`runtimeAssaultCronExpression`** within **`AssaultProperties`** class).

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [ ] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -3,10 +3,12 @@
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
+- https://github.com/codecentric/chaos-monkey-spring-boot/pull/200[#200] Prevent chaos monkey from proxying BeanPostProcessor and ApplicationListener methods (see https://github.com/codecentric/chaos-monkey-spring-boot/issues/199[#199] for more details)
 - https://github.com/codecentric/chaos-monkey-spring-boot/pull/210[#210] Eliminate potential name clash between config props and ENV variables (see https://github.com/codecentric/chaos-monkey-spring-boot/issues/209[#209] for more details)
 
 === Improvements
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
+- https://github.com/codecentric/chaos-monkey-spring-boot/pull/204[#204] [CI] Added release workflow
 
 === New Features
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
@@ -15,5 +17,7 @@
 This release was only possible because of these great humans:
 
 // - https://github.com/octocat[@octocat]
+- https://github.com/alexfincham[@alexfincham]
+- https://github.com/WtfJoke[@WtfJoke]
 
 Thank you for your support!
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -3,11 +3,10 @@
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
-- https://github.com/codecentric/chaos-monkey-spring-boot/pull/200[#200] Prevent chaos monkey from proxying BeanPostProcessor and ApplicationListener methods (see https://github.com/codecentric/chaos-monkey-spring-boot/issues/199[#199] for more details)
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: chaos-monkey-docs/src/main/asciidoc/changes.adoc [hunk 1]
```diff
@@ -16,7 +15,5 @@
 This release was only possible because of these great humans:
 
 // - https://github.com/octocat[@octocat]
-- https://github.com/alexfincham[@alexfincham]
-- https://github.com/WtfJoke[@WtfJoke]
```
WtfJoke, line 20: These stuff shouldnt be deleted

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/AssaultProperties.java
```diff
@@ -38,62 +37,49 @@
 @NoArgsConstructor
 @ConfigurationProperties(prefix = "chaos.monkey.assaults")
 @Validated
-@EqualsAndHashCode
```
WtfJoke, line 41: Thanks for this boy scout :+1:

WtfJoke (commented):
Hey @alexfincham 

Thank you for another contribution, I appreciate that :clap: 

There seems to have happened a mistake while updating `changes.adoc` as the existing changes have been deleted (they should stay in place). 

Except that I was able to verify your fix and it works. 

So would be great if you can fix the changes.adoc file and then I can approve and merge your PR. :+1:

chaos-monkey-docs/src/main/asciidoc/changes.adoc
alexfincham: Good catch...sorry about that

alexfincham: Good catch...sorry about that

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/AssaultProperties.java
```diff
@@ -38,62 +37,49 @@
...
+  private double memoryFillIncrementFraction = 0.15;
 
-  @Value("${memoryFillTargetFraction : 0.25}")
   @DecimalMax("1.0")
   @DecimalMin("0.01")
-  private double memoryFillTargetFraction;
+  private double memoryFillTargetFraction = 0.25;
 
-  @Value("${runtime.scope.assault.cron.expression:OFF}")
+  @Value("${chaos.monkey.assaults.runtime.scope.assault.cron.expression:OFF}")
```
WtfJoke, line 80: nice catch

WtfJoke (approved):
LGTM 📦

Issue #218, opened 2021-04-13
External dependency Jar file broken in 2.3.9 release


georgeharley:
## Expected Behavior

Using the version 2.3.9 external dependencies Jar (chaos-monkey-spring-boot-2.3.9-jar-with-dependencies.jar) to introduce chaos attacks to an existing executable Spring Boot Jar should work [as described in the documentation](https://codecentric.github.io/chaos-monkey-spring-boot/latest/#_external_dependency_jar_file).

## Current Behavior

With the 2.3.9 external dependencies Jar there is no evidence that the target Spring Boot application is affected in any way. There is no "ready to do evil!" message in the application logs and the configured attacks do not occur.

## Possible Solution

The problem appears to be that the 2.3.9 external dependencies Jar has been built with an incorrect **META-INF/spring.factories** file. It ought to pull in the class `de.codecentric.spring.boot.chaos.monkey.configuration.ChaosMonkeyConfiguration` to the Spring Boot autoconfiguration but instead that crucial configuration class is not actualy mentioned. Instead, the **META-INF/spring.factories** file references thousands of Spring classes. 

## Steps to Reproduce

Try following the steps in [section 2.2](https://codecentric.github.io/chaos-monkey-spring-boot/latest/#_external_dependency_jar_file) of the official documentation using **chaos-monkey-spring-boot-2.3.9-jar-with-dependencies.jar**. The steps will fail with no sign of any CM4SB activity. 

Try replacing the **META-INF/spring.factories** file in the external dependencies Jar with the one contained in the 2.3.2 external dependencies Jar and then repeat the steps: things should now work.

Note: the **META-INF/spring-configuration-metadata.json** file included in the 2.3.9 external dependencies Jar also seems to be incorrect. While it does not have any bearing on the problem described in this issue it is perhaps further evidence that something odd happened when **chaos-monkey-spring-boot-2.3.9-jar-with-dependencies.jar** was assembled?

WtfJoke:
Thanks @georgeharley for your detailed report.

I made some additional tests during the last few evenings. It wasnt happening with spring boot 2.4.4 and 2.3.2. 

On my windows machine with 2.3.9 they were two spring.factories file in the jar. The one you mention (which is the default of spring boot) and the chaos monkey one, however on my linux machine I had only the spring boot factories file. 

Although it seems likely, I dont think this is a build issue as we follow the documented approach as far as I can tell.

[...]

Issue #212, opened 2021-03-29
Interested in ability to enable/disable chaos via feature flags?


checketts:
Currently you can configure CM4SB via properties and acutator endpoints. However, I would like to contribute code that would leverage the Unleash feature switch library to enable/disable chaos.

For example, the workflow I want to use is this:

- Unleash will be an optional dep for CM4SB
- Unleash configuration will check if the Unleash class is on the classpath
- If the user enables a feature flag (myApp.controllers.chaos) then chaos will happen, then they can turn it off to disable the chaos.

Would a PR offering this support be welcome?

checketts:
@WtfJoke Sorry to bump this issue. But I would really love to contribute a PR for this if it isn't too out of scope for this project. I'm happy to clarify my usecase if you like.

Here are some benefits why I prefer feature flags over actuator endpoints:

* Since the service pulls in their flag states, we can be sure that each system in a cluster has the same configuration
* Using feature flags I can enable chaos for a subset of my requests. Specifically, I can enable chaos for myself without impacting others
* The UI for enabling/disable chaos settings is a bit more straight forward since it largely can just be a toggle that is turned on or off.

![image](https://user-images.githubusercontent.com/416063/114650153-f4289c00-9c9e-11eb-9b41-b59942ef49ef.png)

WtfJoke:
Hi @checketts 

Sorry for the late reply, currently I'm quiet occupied :see_no_evil: 

Although I like the idea and can see a benefit, I'm a bit worried about the maintenance efforts after a potential PR is merged. And we should then probably support other feature flag libraries such as Togglz. 



>     * Using feature flags I can enable chaos for a subset of my requests. Specifically, I can enable chaos for myself without impacting others

Can you eloborate a bit more how that would work?

checketts:
I completely understand concerns around scope increase/maintenance. In fact the way I would approach it would abstract the logic so togglz (or commercial offerings like Launch Darkly) could easily be plugged in.

So the way Unleash works (and I assume most modern feature switch services do as well) is to include a 'context' when checking if a toggle is enabled. That context can include data like the user's ID, or other metadata like roles or preferred languages, you can add whatever you like.

Essentially the call looks like this, the `isEnabled` call uses a `ContextProvider` to lookup the user details:

```java
if(featureService.isEnabled("includeChaos")) {
  //add chaos to request
}
``` 

That allows approaches like enabling chaos for specific users (testers or automation) or other approaches.

[...]

PR #220, opened 2021-04-17
Add ChaosToggle interface and update tests to account for new the ToggleType enum being passed

checketts:
**What**: This PR addresses #212 : Adding the ability to use feature toggles to enable/disable assaults

**Why**: By allowing feature toggles to control turning on and off chaos monkey attacks, we can treat it like another feature and enable it at a per-user level.

**How**: This PR introduces a `ChaosToggles` interface and a `DefaultChaosToggles` implementation that is hardcoded to 'enabled'.

Also the `ChaosToggles` interface includes a `nameMapper` method so that the user can control how method attacks are grouped into toggle names.

**Checklist**:

- [x] Documentation added
- [x] Changelog updated
- [x] Tests added
- [x] Ready to be merged

**Notes**:

* The current approach also includes an `optional` dependency on Unleash with a very simple implementation of an `UnleashChaosToggles` that uses `Unleash` for checking toggles. I'll be using that to play with how it behaves and iterate on useful defaults
* To use the `nameMapper`, the `isEnabled` method signature now includes the `MetricType` to help group the various toggle checks.

This PR is a Work In Progress and will evolve based on feedback


chaos-monkey-spring-boot/pom.xml
```diff
@@ -126,6 +126,14 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>no.finn.unleash</groupId>
+            <artifactId>unleash-client-java</artifactId>
+            <version>4.2.1</version>
+            <optional>true</optional>
```
checketts, line 133: Note this is an `optional` dep. So users of the library won't get Unleash automatically. This similar approach is used by the Micrometer project.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScope.java
```diff
@@ -109,8 +113,9 @@ private boolean isTrouble() {
         >= chaosMonkeySettings.getAssaultProperties().getLevel();
   }
 
-  private boolean isEnabled() {
-    return this.chaosMonkeySettings.getChaosMonkeyProperties().isEnabled();
+  private boolean isEnabled(MetricType type, String name) {
+    return this.chaosMonkeySettings.getChaosMonkeyProperties().isEnabled()
+        && chaosToggles.isEnabled(chaosToggles.nameMapper(type, name));
```
checketts, line 118: This is the key change: when we check if Chaos Monkey is enabled, we also check that the corresponding toggle is enabled.

The default implementation is `true` to existing users won't need to change.

Note that we are now passing in the MetricType so we can include that in the `nameMapper` and use it to group the toggle. I don't love that we are reusing the MetricType enum, since we aren't using it for metrics purposes, but I like that it does mostly align.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
```diff
@@ -135,8 +140,30 @@ public MemoryAssault memoryAssault() {
...
+    @ConditionalOnMissingBean
+    @ConditionalOnBean(Unleash.class)
+    public ChaosToggles unleashChaosToggles(
+        Unleash unleash, ChaosMonkeyProperties chaosMonkeyProperties) {
+      return new UnleashChaosToggles(chaosMonkeyProperties, unleash);
+    }
+  }
+
+  @Bean
+  @ConditionalOnMissingBean(ChaosToggles.class)
```
checketts, line 164: By using `ConditionalOnMissingBean` we provide a default implementation that mimics the current behavior, while allowing the user to override with others.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/ChaosToggles.java
```diff
@@ -0,0 +1,10 @@
+package de.codecentric.spring.boot.chaos.monkey.configuration.toggles;
+
+import de.codecentric.spring.boot.chaos.monkey.component.MetricType;
+
+public interface ChaosToggles {
```
checketts, line 5: Here is the key interface that allows this feature

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/UnleashChaosToggles.java
```diff
@@ -0,0 +1,29 @@
...
+    this.unleash = unleash;
+  }
+
+  @Override
+  public String nameMapper(MetricType type, String name) {
+    if (type == null) {
+      return chaosMonkeyProperties.getTogglePrefix() + ".all";
+    }
+
+    return chaosMonkeyProperties.getTogglePrefix() + "." + type.getMetricName();
```
checketts, line 22: Here you can see that with Unleash the toggles get grouped by the metric type. So by default, if Chaos Monkey if running and enabled the following toggles would be getting checked:

* chaos.monkey.all
* chaos.monkey.controller
* chaos.monkey.restController
* chaos.monkey.repository
* chaos.monkey.service
* chaos.monkey.component

Put with the name mapping a user could override it and divide it further based on the name input. They can change the prefix, since it is common to prefix toggle based on and application:

* myApp.chaos.monkey.repository
* myOtherApp.chaos.monkey.repository

Thus allowing easy segmentation of toggles between apps, if that is desired.

checketts:
@WtfJoke Did you get a chance to look at this? Sorry the PR is large. I included a demo app to make it easier for you to inspect the idea, but I'm happy removing that from the PR to keep it focused.

If the idea has merit, I was going to start adding documentation/tests, but wanted to hold off until I had cleared the overall concept with you.

(By the way, I'm happy to be patient on your feedback. I completely understand how busy life can get. So I'm interested in making this PR as simple as possible to help with that)

WtfJoke:
@checketts I havent had a look yet, as im currently taking all my free coding time for this project to fix #218 

About the demo app, we have quiet a couple already. So I would probably either delete it or integrate the crucial parts in one of the existing ones. But for now you can keep it, so I can at least have a look at it before you delete it :)

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ToggleType.java
```diff
@@ -0,0 +1,19 @@
+package de.codecentric.spring.boot.chaos.monkey.component;
+
+public enum ToggleType {
+  CONTROLLER("controller"),
+  RESTCONTROLLER("restController"),
```
WtfJoke, line 5:
```suggestion
  REST_CONTROLLER("restController"),
```

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ToggleType.java
```diff
@@ -0,0 +1,19 @@
+package de.codecentric.spring.boot.chaos.monkey.component;
+
+public enum ToggleType {
```
WtfJoke, line 3: I would like to have this not named Toggle as this enum simply represent the attacked Component Type. 
In chaos monkey terms its more of a `WatchType` (or something similar, as I don't like that name very much either, maybe you have a better idea :smile:) as you define which Components you watch.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/component/ToggleType.java
```diff
@@ -0,0 +1,19 @@
...
+
+  private final String toggleName;
+
+  ToggleType(String toggleName) {
+    this.toggleName = toggleName;
+  }
+
+  public String getToggleName() {
+    return toggleName;
+  }
```
WtfJoke, line 18: This would turn into `getName()`

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
```diff
@@ -142,24 +145,36 @@ public MemoryAssault memoryAssault() {
...
+        allAssaults,
+        publisher(),
+        chaosToggles,
+        chaosToggleNameMapper);
   }
 
   @Configuration
   @ConditionalOnClass(Unleash.class)
+  @ConditionalOnProperty(value = "chaos.monkey.toggle.unleash.enabled", matchIfMissing = true)
   static class UnleashChaosConfiguration {
```
WtfJoke, line 162: Please extract this class into a separate file so ChaosMonkeyConfiguration is kept as small as possible

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
WtfJoke: Here im unsure if we need a DefaultChaosToggleNameMapper at all, as its only important if we have Unleashed enabled.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/DefaultChaosToggleNameMapper.java
```diff
@@ -0,0 +1,21 @@
+package de.codecentric.spring.boot.chaos.monkey.configuration.toggles;
+
+import de.codecentric.spring.boot.chaos.monkey.component.ToggleType;
+import de.codecentric.spring.boot.chaos.monkey.configuration.ChaosMonkeyProperties;
+
+public class DefaultChaosToggleNameMapper implements ChaosToggleNameMapper {
+  protected final ChaosMonkeyProperties chaosMonkeyProperties;
+
+  public DefaultChaosToggleNameMapper(ChaosMonkeyProperties chaosMonkeyProperties) {
```
WtfJoke, line 9: I think its sufficient if we pass togglePrefix to the class as the other properties arent really necessary for mapping the name

demo-apps/chaos-monkey-demo-app-unleash-toggles/src/main/java/com/example/chaos/monkey/chaosdemo/ChaosDemoApplication.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Bean
+  public UnleashContextProvider unleashContextProvider() {
+    return () -> {
+      UnleashContext.Builder context = new UnleashContext.Builder();
+
+      try {
+        // This isn't a true production approach. Spring Security would typically be used instead.
+        // This is merely a basic 'fake user' for demo/testing
+        HttpServletRequest curRequest =
```
WtfJoke, line 46: Wouldn't it make sense to just use Spring Security in that demo app? I prefer real world use cases (as long as it doesnt add too much of complexity so that the purpose of the demo is lost)

I think the demo app need some polish (the casting isnt that nice, and the swallow of the exception etc..) and I think there too much duplicated tests (which most likely dont get any value) in that demo project.

demo-apps/chaos-monkey-demo-app-unleash-toggles/src/main/java/com/example/chaos/monkey/chaosdemo/MyAppToggleMapper.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class MyAppToggleMapper extends DefaultChaosToggleNameMapper {
+  public MyAppToggleMapper(ChaosMonkeyProperties chaosMonkeyProperties) {
+    super(chaosMonkeyProperties);
+  }
+
+  @Override
+  public String mapName(ToggleType type, String name) {
+    if (type.equals(ToggleType.CONTROLLER) && name.toLowerCase().contains("hello")) {
+      return this.chaosMonkeyProperties.getTogglePrefix() + ".howdy";
```
WtfJoke, line 15: :cowboy_hat_face: :horse:

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/component/ChaosMonkeyRequestScopeTest.java
```diff
@@ -64,14 +66,16 @@ void setUpCommon() {
             chaosMonkeySettings,
             Arrays.asList(latencyAssault, exceptionAssault),
             Collections.emptyList(),
-            metricEventPublisherMock);
+            metricEventPublisherMock,
+            new DefaultChaosToggles(),
+            new DefaultChaosToggleNameMapper(chaosMonkeySettings.getChaosMonkeyProperties()));
```
WtfJoke, line 71:
```suggestion
            new DefaultChaosToggleNameMapper(chaosMonkeyProperties);
```

WtfJoke (commented):
I've gone through the most of the changes and left some comments.
Most likely I would keep the demo app, but I'm a bit unsure about it :sweat_smile: 

Would love if @fletchgqc could have a look at the PR as well and share his thoughts.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/ChaosMonkeyConfiguration.java
checketts: We need a default since the calling code to check the toggle uses it: 
```java
chaosToggles.isEnabled(chaosToggleNameMapper.mapName(type, name))
```

If a user was to create their own `ChaosToggle` implementation , they would want to breakout by `ChaosTarget` that the default mapper provides.

demo-apps/chaos-monkey-demo-app-unleash-toggles/src/main/java/com/example/chaos/monkey/chaosdemo/ChaosDemoApplication.java
checketts: I've pulled in Spring Security and updated the examples.  I removed some of the tests. Several of those tests, like the ControllerTests aren't exercising any ChaosMonkey behavior.   Should I remove those? 

I'm tempted to update them to trigger chaos behaviors, so they can serve as an example to others, but if that isn't likely to be useful in actual use cases, I'm happy to delete those tests so they don't distract from the others.

checketts:
Thanks for the review! I'll start adding more tests and documentation to explain usecases and approaches.

checketts:
OK. I've made the recommended updates and added the tests, documentation, and changelog

fletchgqc:
Hi, I had a look. So far a good effort has been put in here and it's a nice looking PR.

I get the basic idea with Unleashed, it looks like a good library. But by adding that, we take on a lot of complexity. Two configuration mechanisms, two places to look when things aren't working, more complicated documentation and harder to get started. And more maintenance for us. I understand that by default Unleashed is simply deactivated, which helps, but it doesn't totally remove the complexity.

In exchange for that complexity, we need to be adding corresponding value. Could you explain some realistic use cases which would be possible with this addition?

Another question: if you have two pods running in Kubernetes, is it possible to address them with Unleashed? This is a problem we currently have with actuator - you can't direct the requests to a particular pod.

Would integration with Spring Boot Admin, which is under consideration, be another way to fulfill the use cases which Unleashed fulfills? This offers a kind of centralised management of your services (behind the scenes via actuator).

A technical point: Why is Spring Security needed?

chaos-monkey-dependencies/pom.xml
```diff
@@ -72,6 +72,11 @@
                 <artifactId>spring-boot-starter-validation</artifactId>
                 <version>${spring-boot.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-starter-security</artifactId>
+                <version>${spring-boot.version}</version>
+            </dependency>
```
WtfJoke, line 79: Spring Security should be only needed in the demo app and added only there as dependency.

If thats not possible, im sorry for bringin this up, then I would preferr the initial approach 🙈

demo-apps/chaos-monkey-demo-app-unleash-toggles/src/main/java/com/example/chaos/monkey/toggledemo/ChaosDemoApplication.java
```diff
@@ -0,0 +1,121 @@
...
+
+      return context.build();
+    };
+  }
+
+  @Bean
+  public Unleash unleash(UnleashContextProvider contextProvider) {
+    UserAwareFakeUnleash unleash = new UserAwareFakeUnleash(contextProvider);
+
+    //        unleash.enable("chaos.monkey.controller");
```
WtfJoke, line 65:
```suggestion

```

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/DefaultChaosToggleNameMapperTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import de.codecentric.spring.boot.chaos.monkey.component.ChaosTarget;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class DefaultChaosToggleNameMapperTest {
+
+  DefaultChaosToggleNameMapper sut;
```
WtfJoke, line 11:
```suggestion
  private DefaultChaosToggleNameMapper sut;
```

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/DefaultChaosTogglesTest.java
```diff
@@ -0,0 +1,22 @@
+package de.codecentric.spring.boot.chaos.monkey.configuration.toggles;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class DefaultChaosTogglesTest {
+
+  DefaultChaosToggles sut;
```
WtfJoke, line 10:
```suggestion
  private DefaultChaosToggles sut;
```

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/UnleashChaosTogglesTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import no.finn.unleash.FakeUnleash;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class UnleashChaosTogglesTest {
+  UnleashChaosToggles sut;
+  FakeUnleash fakeUnleash;
```
WtfJoke, line 11:
```suggestion
  private UnleashChaosToggles sut;
  private FakeUnleash fakeUnleash;
```

pom.xml
```diff
@@ -38,6 +38,7 @@
     <spring-boot.version>2.4.5</spring-boot.version>
     <spring-cloud.version>2020.0.2</spring-cloud.version>
     <spring.version>5.3.6</spring.version>
+    <spring.security.version>5.3.6.RELEASE</spring.security.version>
```
WtfJoke, line 41:
```suggestion
    <spring.security.version>5.4.6</spring.security.version>
```

According to https://docs.spring.io/spring-boot/docs/2.4.5/reference/html/appendix-dependency-versions.html it should be 5.4.6

checketts:
> In exchange for that complexity, we need to be adding corresponding value. Could you explain some realistic use cases which would be possible with this addition?

The use cases I foresee using is to enable Chaos Monkey profiles and properties in all my lanes, however I be leaving the toggles off. That will remove one part changing in the lanes.

We already use feature flags as part of our continuous deployment strategy. As we add new feature or even patch old code, we wrap those items in feature toggles. I suspect anyone adopting Unleash would be doing a similar workflow.

Feature toggle evaluation happens on a per-request basis so with this change I can now 

* Enable chaos for specific users (test workloads or downstream clients I have convinced to expect chaos responses in production traffic)
* Enable chaos using the same mechanism we use for feature switching
* Using it as a kill switch (I'll come back to this below)

> Another question: if you have two pods running in Kubernetes, is it possible to address them with Unleashed? This is a problem we currently have with actuator - you can't direct the requests to a particular pod.

This use case is addressed by Unleash. Since the servers call back to Unleash we can be certain all servers have pulled in the feature config. And that feature config allow enabling a toggle based on parameter such as host ID. Configuring it to use a pod's ID would be very simple.

> Would integration with Spring Boot Admin, which is under consideration, be another way to fulfill the use cases which Unleashed fulfills? This offers a kind of centralised management of your services (behind the scenes via actuator).

Spring Boot Admin could address the config question, but won't solve the Kubernetes use case mentioned above and also won't allow for the per-user/request evaluation.

> A technical point: Why is Spring Security needed?

Spring Security is being used just for the toggle demo app to demonstrate chaos being enabled for one user but not another. I could move that from `dependencyMangement` into the demo app if that is preferred.

chaos-monkey-dependencies/pom.xml
checketts: You are correct, it is only used for the demo. Should I move it into that demo app, or is leaving it here OK?

demo-apps/chaos-monkey-demo-app-unleash-toggles/src/main/java/com/example/chaos/monkey/toggledemo/ChaosDemoApplication.java
checketts: I left this comment as a way to demonstrate how someone could toggle a feature without having to run an actual Unleash server. Should I remove it or just add additional comments explaining its use?

checketts:
Sorry for all the mini-commits when I accepted your suggestions. I hadn't realized you could add them to a batch. Let me know if you want me to rebase them into the main commit to clean up the history.

chaos-monkey-dependencies/pom.xml
WtfJoke: I'll preferr if its moved to the demo app.

checketts:
@WtfJoke I removed spring-security as a top level dependency

fletchgqc:
I think the use cases are interesting. Whilst I personally don't have a use for switching CMSB on per-request at the moment, it's an interesting idea to be able to activate it via header, for example. These kind of flows might be particularly important for people using it in production. So I agree with integrating it. I like that the documentation explains the default status pretty well.

We are going to add a new watcher soon - rest template and webclient. We'll need to add default toggles for them, when that is done.

Speaking of default status... for some insane historical reason, latency on services is enabled by default in CMSB. It's rather confusing. Does anyone volunteer to create a PR to turn everything off by default?

fletchgqc:
Another thought about defaults and configuration mechanisms. Would it be more user-friendly, instead of a layered on/off configuration, that you determine via configuration whether toggles or actuator will be used? I.e. when you turn on toggles, it's irrelevant what is configured via actuator?

fletchgqc:
Speaking of default status... for some insane historical reason, latency on services is enabled by default in CMSB. It's rather confusing. I would be happy if someone created a PR to turn everything off by default :-)

checketts:
> We are going to add a new watcher soon - rest template and webclient. We'll need to add default toggles for them, when that is done

I'm very eager for the RestTemplate and WebClient code. I Cherrypicked some of the code from an example and was very pleased with how it behaved. I'll be glad to help contribute to those PRs.

> Would it be more user-friendly, instead of a layered on/off configuration, that you determine via configuration whether toggles or actuator will be used? 

I did a spike with that since with Unleash toggles you can set a 'variant' (essentially return a string instead of a boolean). It was fun to be able to configure specific toggles, but in reality I found that I didn't need custom config per toggle and if I did, configuring it via the properties (with the toggle off) was sufficient.

I haven't had much experience with actuator experience though, so my perspective isn't exhaustive 9and still quite new to this).

>  ...latency on services is enabled by default in CMSB. I would be happy if someone created a PR to turn everything off by default :-)

I agreed and will be happy to tackle that once we get this PR taken care of.

WtfJoke (commented):
I've gone through it now more thoroughly. This should be the last iteration in order to get this merged (at least I hope that).

Hope you dont get shocked when you see the amount of comments I left (but most of them are rather small) :sweat_smile: 

I am going to test this in a separate project to verify the newly added toggles docs.

chaos-monkey-spring-boot/pom.xml
```diff
@@ -120,6 +120,14 @@
...
+        </dependency>
```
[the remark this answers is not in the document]
checketts: OK. I think I get it now. We didn't put the spring-security dependencies in the `chaos-monkey-depedencies`, because Chaos Monkey isn't dependent on it, but Unleash is.

checketts:
Thanks for the comments @WtfJoke . I appreciate the good feedback. I've incorporated all your recommendations.

demo-apps/chaos-monkey-demo-app-ext-jar/pom.xml
```diff
@@ -44,7 +44,7 @@
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-registry-prometheus</artifactId>
-            <version>1.6.6</version>
+            <version>${micrometer.version}</version>
```
WtfJoke, line 47: <3

WtfJoke:
> Let me know if you want me to rebase them into the main commit to clean up the history.

You can, but I'll most likely just squash them

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/ChaosToggleNameMapper.java
```diff
@@ -0,0 +1,20 @@
+package de.codecentric.spring.boot.chaos.monkey.configuration.toggles;
+
+import de.codecentric.spring.boot.chaos.monkey.component.ChaosTarget;
+
+/**
+ * A way to map individual ChaosTargets (controllers, repositories, etc) and the corresponding
+ * method. Implementations will can make the name to toggle mapping as coarse or as detailed as
```
WtfJoke, line 7:
```suggestion
 * method. Implementations can make the name to toggle mapping as coarse or as detailed as
```

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/configuration/toggles/ChaosToggleNameMapper.java
```diff
@@ -0,0 +1,20 @@
+package de.codecentric.spring.boot.chaos.monkey.configuration.toggles;
+
+import de.codecentric.spring.boot.chaos.monkey.component.ChaosTarget;
+
+/**
+ * A way to map individual ChaosTargets (controllers, repositories, etc) and the corresponding
+ * method. Implementations can make the name to toggle mapping as coarse or as detailed as
+ * desired.
```
WtfJoke, line 8:
```suggestion
 * method. Implementations can make the name to toggle mapping as coarse or as detailed as desired.
```

WtfJoke (approved):
GreetingServiceTest.java  is still there (most likekly just overlooked by the sheer amount of comments), but else it looks good. I fixed a typo in the newly added javadoc.

eg. LGTM 🐟 

Thanks a lot @checketts for your contribution and your invested time :)

WtfJoke:
Forgot to mention that my test using unleash and one of the new toggles worked as expected 👍

Issue #93, opened 2019-08-26
Watcher for outgoing Rest-Calls (Template)


ChristianLangmann:
Instead of watching components that implement a function and assaulting it, watch for outgoing calls (using RestTemplate) and assault this.

haridevelops:
what exactly needs to be fixed ? can you elaborate ?

elijah-pl:
I think the goal should be achieved by writing an interceptor that will be injected into the existing `RestTemplate` bean and randomly return different responses (timeout, internal server error, ok etc.) or introduce delays.

```java
@Bean
@Primary // <- Not sure if this will be needed, probably not
@ConditionalOnBean(RestTemplate.class)
@ConditionalOnProperty(..) // <- Check if RestTemplate assaulting is enabled
public RestTemplate restTemplate(RestTemplate restTemplate) {
    restTemplate.getInterceptors().add(new ChaosMonkeyInterceptor(..));
    return restTemplate;
}
```
If @maiksensi is OK with this solution, I would like to prepare a MR for this.

maiksensi:
Hey @haridevelops and @elijah-pl,

thanks for your interest in this issue. It is highly appreciated. Currently, Chaos Monkey for Spring Boot (CMSB) is able to assault several annotations one can use. Such as `@Controller`, or `@Service`. Assaulting outgoing calls (where `@RestTemplate` is used) are not part of the possible classes that can be assaulted. 

However, so far we did not discuss the exact scope of that feature. 

@elijah-pl your propsal sounds interesting. I wonder whether it is possible to match "all possible" use cases with that? Because `RestTemplate` is not an Annotation. `RestTemplate` can be used with or without Autowiring. I take it that the pseudo code above would only work for cases where you would do something like this: 

```
....
@Autowired
RestTemplate restTemplate;
...
```

What about cases where a `RestTemplate` is instantiated like this: 

```
....
public void someMethod() {
    RestTemplate restTemplate = new RestTemplate();
}
```

haridevelops:
We can create something like RestCommonClient.java which can be inherited which class wants to do a rest call.

So RestTemplate will a class level variable of RestCommonClient. hence can be utilised all across.

For your case like new RestTemplate() scenario, 
There won't be a much of code change than referencing the right object from parent .

haridevelops:
Like how we add an Error handler for a  RestTemplate.

All the exceptions and intercepting can be done at one place which can utilized all across.

[...]

PR #229, opened 2021-05-28
#93 Add RestTemplate and WebClient watchers

tuxdevelop:
**What**: Implemented issue #93 - providing watchers for RestTemplate and WebClient

**Why**: feature request #93 

**How**: providing BeanPostProcessors to inject a interceptor (RestTemplate) or a filter (WebClient) which act as a watcher to provide exception and latency assaults

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [x] Tests added
      
- [x] Ready to be merged


chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplatePostProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.web.client.RestTemplate;
+
+/** @author Marcel Becker */
+public class ChaosMonkeyRestTemplatePostProcessor implements BeanPostProcessor {
+
+  private final ChaosMonkeyRestTemplateWatcher interceptor;
+
+  public ChaosMonkeyRestTemplatePostProcessor(final ChaosMonkeyRestTemplateWatcher interceptor) {
```
checketts, line 12: Have you considered using a [RestTemplateCustomizer](https://www.baeldung.com/spring-rest-template-builder)? I think the is a preferred mechanism for modifying RestTemplate/WebClient

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcher.java
```diff
@@ -0,0 +1,103 @@
...
+      final HttpRequest httpRequest,
+      byte[] bytes,
+      ClientHttpRequestExecution clientHttpRequestExecution)
+      throws IOException {
+    ClientHttpResponse response;
+    response = clientHttpRequestExecution.execute(httpRequest, bytes);
+    if (watcherProperties.isRestTemplate()) {
+      try {
+        chaosMonkeyRequestScope.callChaosMonkey(
+            ChaosTarget.REST_TEMPLATE, this.getClass().getSimpleName());
```
checketts, line 47: The class name won't change in a meaningful way. I would recommend naming the toggle based on the host. That way users can target the chaos against specific dependencies.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientPostProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+    this.filter = filter;
+  }
+
+  @Override
+  public Object postProcessBeforeInitialization(final Object bean, final String beanName)
+      throws BeansException {
+    final Object target;
+    if (bean instanceof WebClient) {
+      // create a copy of WebClient whose settings are replicated from the current WebClient.
+      target = ((WebClient) bean).mutate().filter(filter).build();
```
checketts, line 22: Similarly, use the [WebClientCusomizer](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/reactive/function/client/WebClientCustomizer.html) to add your filter

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public Mono<ClientResponse> filter(
+      ClientRequest clientRequest, ExchangeFunction exchangeFunction) {
+    Mono<ClientResponse> response;
+    response = exchangeFunction.exchange(clientRequest);
+    if (watcherProperties.isWebClient()) {
+      try {
+        chaosMonkeyRequestScope.callChaosMonkey(
+            ChaosTarget.WEB_CLIENT, this.getClass().getSimpleName());
```
checketts, line 39: Change the `this.getClass().getSimpleName()` to the request's host:

Here is a snippet I'm using in my WebClient (similar to Micrometer approach for metrics for WebClient)
```
    private fun getClientName(request: ClientRequest): String {
        return request.url().host ?: return "unknown_host"
    }
```

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
```diff
@@ -0,0 +1,75 @@
...
+      } catch (final Exception exception) {
+        try {
+          if (exception.getClass().equals(assaultProperties.getException().getExceptionClass())) {
+            response = Mono.just(ErrorClientResponse.getResponse());
```
checketts, line 43: In addition to error responses, is there support for delays? Here is a snippet on how WebClient delay is done:

```
            if(latency > 0) {
                log.warn { "Chaos enabled for ${clientRequest.url()} adding latency of ${latency}ms" }
                return next.exchange(clientRequest).delayElement(Duration.ofMillis(latency))
            }
```

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
```diff
@@ -0,0 +1,75 @@
...
+      HttpStatus.UNAUTHORIZED,
+      HttpStatus.NOT_FOUND,
+    };
+
+    static final String ERROR_BODY =
+        "{\"error\": \"This is a Chaos Monkey for Spring Boot generated failure\"}";
+
+    private static ClientResponse getResponse() {
+      return ClientResponse.create(
+              ERROR_STATUS_CODES[new Random().nextInt(ERROR_STATUS_CODES.length)])
```
checketts, line 70: I'm not certain that making this random is preferred. Perhaps making it configurable and if more than one is configured, randomization will be used than?

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcherIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+      classes = {ChaosDemoApplication.class})
+  @ActiveProfiles("chaos-monkey")
+  static class LatencyAssaultIntegrationTest {
+
+    @Autowired private DemoRestTemplateService demoRestTemplateService;
+
+    @Test
+    public void testRestTemplateLatencyAssault() {
+      try {
+        this.demoRestTemplateService.callWithRestTemplate();
```
checketts, line 83: `assertThatThrownBy` (from [AssertJ](https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html)) is a cleaner way to check that an exception occurs

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcherIntegrationTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Autowired private DemoWebClientService demoRestTemplateService;
+
+    @Test
+    public void testWebClientLatencyAssault() {
+      try {
+        this.demoRestTemplateService.callWithWebClient();
+        fail("No ReadTimeoutException occurred!");
+      } catch (Exception ex) {
+        BDDAssertions.then(ex.getCause() instanceof ReadTimeoutException).isTrue();
+        log.debug("exception caught, everything is fine!", ex);
```
checketts, line 65: Same here, switch to `assertThatThrownBy` and remove debugging

checketts (commented):
Thanks for the PR! I just had a few recommendations around the toggle names and the exception assertions.

WtfJoke (commented):
Looks good in general, I left some comments here and there and distributed some :+1: of the existing review. 

Dont forget to update the changelog entry ;) (it would be great if you could fix the formatting issue it currently has) 

Havent tested it in a separate project yet (will do that before we merge it)

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplatePostProcessor.java
tuxdevelop: Hi, I also thought about the customizer, but there is one big challenge/issue: Spring Boot just applies the RestTemplateCustomizers to RestTemplates which are created via injected RestTemplateBuilder. This approach will miss from my experience 99% of all created RestTemplate beans. 

The issue is also described very well in the article: https://medium.com/@TimvanBaarsen/spring-boot-why-you-should-always-use-the-resttemplatebuilder-to-create-a-resttemplate-instance-d5a44ebad9e9 

My suggestion would be to use the bean post processor approach in combination with the customizer.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcher.java
tuxdevelop: I agree, I will change it to  `getURI().toUrl().toString()`

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
tuxdevelop: I agree, I will change it

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
tuxdevelop: I was thinking about this as well, but I had d sync upfront with @WtfJoke and @fletchgqc and we came to the conclusion, that the assaults should not be included in the watchers. The latency assault is done via  `ChaosMonkeyRequestScope`. 
The handling of the exception is just for the case, that anything else in the filter chain throws an exception except the 'ChaosMonkeyRequestScope'.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
tuxdevelop: Yes, also here I synced with @WtfJoke and we came the conclusion, that we do not have a structure how to handle this kind of properties and we would approach this, if someone will will request it. We can fallback to a default 500 for now?

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcherIntegrationTest.java
tuxdevelop: agree.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcherIntegrationTest.java
tuxdevelop: agree.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcher.java
```diff
@@ -0,0 +1,103 @@
...
+      HttpStatus.INTERNAL_SERVER_ERROR.value(),
+      HttpStatus.BAD_REQUEST.value(),
+      HttpStatus.FORBIDDEN.value(),
+      HttpStatus.UNAUTHORIZED.value(),
+      HttpStatus.NOT_FOUND.value(),
+    };
+
+    @Override
+    public int getRawStatusCode() {
+      return ERROR_STATUS_CODES[new Random().nextInt(ERROR_STATUS_CODES.length)];
```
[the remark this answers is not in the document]
tuxdevelop: For now, I would suggest to go with 2. and use just 500.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/demo/chaos/monkey/service/DemoRestTemplateService.java
```diff
@@ -0,0 +1,19 @@
...
+public class DemoRestTemplateService {
+
+  private final RestTemplate restTemplate;
+
+  public DemoRestTemplateService(RestTemplate restTemplate) {
+    this.restTemplate = restTemplate;
+  }
+
+  public ResponseEntity<?> callWithRestTemplate() {
+    return this.restTemplate.getForEntity("https://www.google.de", String.class);
```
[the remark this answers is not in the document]
tuxdevelop: agree :D

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcher.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public void close() {}
+
+    @Override
+    public InputStream getBody() throws IOException {
+      ByteArrayInputStream inputStream =
+          new ByteArrayInputStream(ERROR_BODY.getBytes(StandardCharsets.UTF_8));
+      inputStream.close();
```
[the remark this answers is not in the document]
tuxdevelop: agree.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientPostProcessor.java
tuxdevelop: Hi, also here we will have a limitation, that just `WebClients` which are created via injected `WebClient.Builder()` will be customized. In this case we do not have the flexibility, like with `RestTemplate` to combine both approaches, because we cannot control to add the `Filter` only once. @WtfJoke any thoughts how to approach this? The BeanPostProcessor approach is not nice, but safe.

checketts: > The BeanPostProcessor approach is not nice, but safe.

Actually this approach misses all `WebClients` that are created by the injected builder. Since in my projects we don't pass around individual `WebClient` beans, we instead use composition in the services. 

For example we have an `EmailFacade` and a `UserFacade` the each contains their own `WebClient`s that were created from the builder.

I would recommend using the customizer, and if you want to include the bean post processor, make the filter behave like a `OncePerRequestFilter` by setting an attrbitute on the request or response and not firing if it has already fired. (I couldn't find a `WebClient` example for you, but the concept is used in HttpFilters and Micrometer MeterFilters)

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
checketts: I wonder if we should change the `.callChaosMonkey()` method to take an Object instead of a String so the name mapper could leverage the objects passed in instead of parsing strings.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
checketts: I'm not sure I understood. Are you saying there isn't a way to leverage the assault config in this location? Or are you saying we don't want to mix the watcher and assault concerns?

I'm a bit new to the project and haven't internalized the difference between watcher and assault.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
checketts: Sounds good.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientPostProcessor.java
WtfJoke: So did I get this right. If we use BeanPostProcessor all injected Webclients will be assaulted (but not the one by injected builders). And if we use WebclientCustomizer its vice versa.


> The BeanPostProcessor approach is not nice, but safe.

> Actually this approach misses all `WebClients` that are created by the injected builder.

Arent these statements contradicting? Or am I missing a point here? :sweat_smile: 


So the difficultiy is to assault both but dont assault it twice.

tuxdevelop: Hi @checketts,
I understand your issue, I did not consider this approach. After a sync with @WtfJoke, we are going to implement both approaches and use the `OncePerRequestFilter`pattern to avoid multi assaults. Thank you for the feedback.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyWebClientWatcher.java
tuxdevelop: Hi,
correct, the current status quo is not to mix assault and watchers. @WtfJoke correct me if I am wrong.

chaos-monkey-spring-boot/src/test/java/de/codecentric/spring/boot/chaos/monkey/watcher/outgoing/ChaosMonkeyRestTemplateWatcherIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    public void testRestTemplateExceptionAssault() {
+      try {
+        this.demoRestTemplateService.callWithRestTemplate();
+        fail("No HttpClientErrorException occurred!");
+      } catch (HttpStatusCodeException ex) {
+        String message = ex.getMessage();
+        assertThat(message)
+            .isEqualTo(ex.getRawStatusCode() + " " + ERROR_TEXT + ": [" + ERROR_BODY + "]");
+      }
+    }
```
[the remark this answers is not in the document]
tuxdevelop: I can replace this, when we just return one error status code.

chaos-monkey-docs/src/main/asciidoc/watcher.adoc
```diff
@@ -14,6 +14,13 @@ With the help of https://docs.spring.io/spring/docs/current/spring-framework-ref
 execution of a _public_ method and will either not execute any action or start one of its assaults. You can customize the behave by
 <<configuration,configuration>>.
 
+Besides the _AOP_ based watchers, Chaos Monkey for Spring Boot also provides watchers for:
+
+- RestTemplate
+- WebClient
+
+The watchers will inject exception and latency based assaults for all _RestTemplate_ and _WebClient_ beans in the Spring Context based on the given assault configuration.
```
checketts, line 22: We might want to add that the the builders are also instrumented and a warning that the clients instantiated and used within  a class (`WebClient.create() and `new RestTemplate`) are not instrumented but can be if the filter is added manually.

tuxdevelop: done

checketts (dismissed):

WtfJoke (approved):
I made some smaller changes (I extracted the ChaosMonkeyRestTemplate and WebClient Configuration into separate classes). 

Now LGTM 🥳

WtfJoke (approved):

Issue #239, opened 2021-07-09
Add CPU Assault


WtfJoke:
It would be great to add a CPU assault which puts load on your cpu to a certain amount/time.

Calculating fibonacci would probably a good way to gain load

PR #241, opened 2021-07-12
Add CPU Assault

F43nd1r:
**What**:
 Add CPU assault (#239)

**Why**:
 New feature

**How**:
 CPU load is generated by spawning busy waiting threads

**Checklist**:

- [x] Documentation added
      
- [x] Changelog updated
      
- [X] Tests added
      
- [x] Ready to be merged


chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/assaults/CpuAssault.java
```diff
@@ -0,0 +1,83 @@
...
+      metricEventPublisher.publishMetricEvent(MetricType.CPU_ASSAULT);
+    }
+    double load = settings.getAssaultProperties().getCpuLoadTargetFraction();
+
+    List<Thread> threads = new ArrayList<>();
+    for (int num = 0; os.getProcessCpuLoad() < load; num++) {
+      Thread thread =
+          new Thread(
+              () -> {
+                while (!Thread.interrupted()) ;
```
checketts, line 50: My understanding is that spawning new threads actually consumes more memory. I wonder if the CPU assault is instead using up memory, then forcing the application to perform garbage collection, which reflects as high CPU load.

I wonder if the thread should be doing some CPU intensive work as well or if this alone is sufficient.

F43nd1r: Threads do consume memory, though not enough to trigger excessive garbage collection in my testing. Nonetheless, reducing the number of threads necessary seems like a good idea, so I added some fibonacci computation, as suggested in the original issue.

WtfJoke (commented):
Looks good, but something is off.
The target fraction doesnt seem to apply on my machine.
I made a test in my sample app and the cpu skyrockets to 95% :D even when setting the fraction to a low value such as "0.1". Having tested on jdk 11 and windows. Could also test on my 🐧 machine.

F43nd1r:
> Looks good, but something is off.
> The target fraction doesnt seem to apply on my machine.
> I made a test in my sample app and the cpu skyrockets to 95% :D even when setting the fraction to a low value such as "0.1". Having tested on jdk 11 and windows. Could also test on my 🐧 machine.

There were several issues with this. The most important one: CPU load is returned for a window, and only updates about twice per second (on my machine, not documented). This meant a lot of threads were spawned before their load was observed.

Fixed issues:
 - not waiting for new data from the os
 - not stopping threads if some other part of the application generates load
 - tests generating most of the load themselves

Persisting issues:
 - monitoring cpu load and managing assault threads generates load, which makes low values (<0.2 on my system) nonsensical
 - target accuracy increases with target value, meaning low values might be overshot by several percent points
 - assault needs considerable ramp up time (up to several seconds)

Are those acceptable shortcomings?

WtfJoke (approved):
LGTM 💻

PR #248, opened 2021-07-22
Document extjar spring version should match

F43nd1r:
**What**:
 Add note about spring boot version requirements to extjar documentation

**Why**:
 Improve documentation

**How**:
Switch asciidoctor build to maven to be able to inject spring boot version

**Checklist**:

- [X] Documentation added
      
- [`N/A`] Changelog updated  
      
- [`N/A`] Tests added  
      
- [ ] Ready to be merged


F43nd1r:
@WtfJoke I'd like to add a note on how to find the correct spring boot version. Will we continue to match the Major.Minor version of Spring Boot?

WtfJoke:
I think we should add that note only in the ext jar part of the documentation. There its important.

Our 2.5.x releases will all build on 2.5.x spring boot releases. However there are patch releases of spring boot which can break compatibility when using extjars (while working fine when adding as dependency).

.github/workflows/releasetag.yml
```diff
@@ -50,16 +50,24 @@ jobs:
...
+          java-version: 11
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Convert changelog to docbook
+        run: ./mvnw asciidoctor:process-asciidoc@output-docbook -pl chaos-monkey-docs
```
WtfJoke, line 66: <3

WtfJoke (approved):
LGTM 📜

PR #255, opened 2021-07-30
Deduplicate documentation

F43nd1r:
**What**:
 Previously, we had three sources of documentation (readme, gh-pages readme and asciidoc), with a lot of content duplicated between them. This brings it down to two (readme + gh-pages readme will have the same content) and moves contents such that no information is duplicated between the two sources.

**Why**:
 Confusing documentation setup, some informations are hard to find.

**How**:
 Change readme and docs, release workflow now also copies readme to gh-pages. Readme now contains meta information like introduction, why and how. asciidoc only contains usage informations.

**Checklist**:

- [X] Documentation added
      
- `N/A` Changelog updated
      
- `N/A` Tests added
      
- [x] Ready to be merged

### DIFF
--- a/.github/workflows/releasetag.yml
+++ b/.github/workflows/releasetag.yml
@@ -26,7 +26,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: Run Maven deploy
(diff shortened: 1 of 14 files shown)


WtfJoke (approved):
LGTM ✏️

Issue #257, opened 2021-07-30
[Kotlin] Document necessity of gradle spring boot plugin


WtfJoke:
The kotlin plugin for spring boot needs to be added for chaos monkey (and a lot of spring related stuff) to work.

It makes the spring classes (such as services, rest-controller etc..) open, so that spring can proxy them (and chaos monkey can attack them)

We should add that to documentation somewhere. 
kotlin("plugin.spring") version "1.5.20"

PR #259, opened 2021-08-13
Document open class requirements

F43nd1r:
**What**:
 Update docs to note the requirement for open classes and recommend kotlin spring plugin.

**Why**:
 #257

**How**:
 Add a new section "Preparing your application"

**Checklist**:

- [X] Documentation added
      
- N/A Changelog updated
      
- N/A Tests added
      
- [X] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/getting-started.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/getting-started.adoc
@@ -99,6 +99,50 @@ CAUTION: Each Chaos Monkey version is built for a specific Spring Boot version, 
 
 It is your decision how you want to integrate and use the CM4SB. There is no difference between the two possibilities. However, it's easier to add it as a dependency (see <<_as_dependency>>), so you don't need to adjust the starting of your application.
 
+=== Preparing your application
+
+If you want chaos monkey to attack your classes they need to be open (=non-final). If you're using java, this is the default. Just don't mark your classes or methods as final. If you're using kotlin, we recommend using the kotlin spring plugin:
+
+[source,xml,subs="verbatim,attributes",role="primary"]
+.Maven
+----
+<plugin>
+    <artifactId>kotlin-maven-plugin</artifactId>
+    <groupId>org.jetbrains.kotlin</groupId>
+    <version>${kotlin.version}</version>
+
+    <configuration>
+        <compilerPlugins>
+            <plugin>spring</plugin>
+        </compilerPlugins>
+    </configuration>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-maven-allopen</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+    </dependencies>
+</plugin>
+----
+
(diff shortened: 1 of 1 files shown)


chaos-monkey-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -99,6 +99,50 @@ CAUTION: Each Chaos Monkey version is built for a specific Spring Boot version,
 
 It is your decision how you want to integrate and use the CM4SB. There is no difference between the two possibilities. However, it's easier to add it as a dependency (see <<_as_dependency>>), so you don't need to adjust the starting of your application.
 
+=== Preparing your application
+
+If you want chaos monkey to attack your classes they need to be open (=non-final). If you're using java, this is the default. Just don't mark your classes or methods as final. If you're using kotlin, we recommend using the kotlin spring plugin:
```
checketts, line 104: I think the language names should be uppercased (Java, Kotlin) Also a short summary describing what the plugin does would help.

> ...we recommend using the Kotlin spring plugin which tells the compiler to make classes with Spring annotations open (`@Controller`, `@Service`, etc.):

checketts (approved):

checketts (approved):

chaos-monkey-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -99,6 +99,50 @@ CAUTION: Each Chaos Monkey version is built for a specific Spring Boot version,
 
 It is your decision how you want to integrate and use the CM4SB. There is no difference between the two possibilities. However, it's easier to add it as a dependency (see <<_as_dependency>>), so you don't need to adjust the starting of your application.
 
+=== Preparing your application
+
+If you want chaos monkey to attack your classes they need to be open (=non-final). If you're using Java, this is the default. Just don't mark your classes or methods as final. If you're using Kotlin, we recommend using the https://kotlinlang.org/docs/all-open-plugin.html#spring-support[Kotlin Spring plugin], which automatically opens Spring annotated classes for you:
```
WtfJoke, line 104: I personally probably would remove the sentence "Just don't mark your classes or methods as final." 
Or put it in braces after this is the default (as long as you don't mark them final).


Thoughts for future:
We might should consider moving this in a (currently not existing) troubleshoot section. 

Then this could be shortened and the Kotlin part could be moved to the beginning: "In case you use Kotlin, make sure you have the Kotlin Spring Plugin enabled, which is responsible for opening Spring annotated classes."

And the java part (where it usually shouldnt be a problem) could come at the end.
"In case of Java, just make sure to not manually mark the assaulted class as final".

F43nd1r: > I personally probably would remove the sentence "Just don't mark your classes or methods as final."
Or put it in braces after "...this is the default (as long as you don't mark them final)". What do you think?

Went for the second one.

> I want to have the docs to read in order that you get ready to do chaos as short as possible

As this is a necessity for kotlin, I don't think it is too misplaced in the setup category. Wouldn't be opposed to a troubleshooting second either.

WtfJoke (approved):
LGTM even with my nitpicking 😅

Issue #82, opened 2019-07-18
KillAppAssault is not working


MelleD:
Hello,

we are using Jetty servlet stack together with Spring webflux.
If I use the `KillAppAssault`, i will get an error by the `Thread.sleep(5000);`

The error is:
```
13:02:16.148] [b34539e37b4747a78b70754a9e1dfb45] [INFO ] [tp1662417126-54] [org.eclipse.jetty.server.session        ] : node0 Stopped scavenging
[13:02:16.151] [b34539e37b4747a78b70754a9e1dfb45] [INFO ] [tp1662417126-54] [o.e.j.s.h.ContextHandler.application    ] : Destroying Spring FrameworkServlet 'dispatcherServlet'
[13:02:16.152] [b34539e37b4747a78b70754a9e1dfb45] [INFO ] [tp1662417126-54] [o.e.jetty.server.handler.ContextHandler ] : Stopped o.s.b.w.e.j.JettyEmbeddedWebAppContext@10f384a2{application,/,[file:///AppData/Local/Temp/jetty-docbase.8177349965831428657.8090/],UNAVAILABLE}
java.lang.InterruptedException: sleep interrupted
	at java.lang.Thread.sleep(Native Method)
	at my.MyResource.query(MyResource.java:71)
	at my.MyResource$$FastClassBySpringCGLIB$$3bcb8e2c.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:749)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:688)
	at my.MyResource$$EnhancerBySpringCGLIB$$491d4ce9.query(<generated>)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #260, opened 2021-08-13
Fix exception in KillAppAssault when using netty

F43nd1r:
**What**:
 Ignore interrupts and continue wait in KillAppAssault

**Why**:
 #82

**How**:
 additional try/catch

**Checklist**:

- N/A Documentation added
      
- [X] Changelog updated
      
- N/A Tests added
      
- [X] Ready to be merged


WtfJoke (approved):
So the change is simply ignoring any interruptedexception?

LGTM

F43nd1r:
> So the change is simply ignoring any interruptedexception?

Exactly - apparently netty interrupts all threads, but we want to wait anyway, so ignore that.

Issue #205, opened 2021-02-23
Should work with Project Lombok`s logger


NicoSe:
To start with chaos engineering I found your project. It look nice to me, but it don´t work as easy as you mensioned it, because in our application I got an NullPointerException. Look the stacktrace below. Hopefully you can fix this over time.

![Screenshot 2021-02-23 184938](https://user-images.githubusercontent.com/50381227/108885290-ec863880-7607-11eb-9a4d-767e0feb995f.png)

Because of that Tomcat could not start and the application shut down. 
I am sure that the problem ist because we use Lombok and its integrated Logger.

WtfJoke:
Hi @NicoSe 

Can you show me an example project, which demonstrates the issue?

Chaos Monkey for Spring Boot is compatible with projects which uses lombok, see https://github.com/fletchgqc/mediator for example. We dont use logger, there, so that might be worth testing.

F43nd1r:
This has nothing to do with lombok. See #261

PR #261, opened 2021-08-13
Exclude GenericFilterBean from proxying

F43nd1r:
**What**:
 Excludes GenericFilterBean from proxying

**Why**:
 GenericFilterBean will throw an exception when proxied due to a final init method. See also #205

**How**:
 Blcklist Pointcut to exclude GenericFilterBean

**Checklist**:

N/A Documentation added
      
- [X] Changelog updated
      
- N/A Tests added
      
- [X] Ready to be merged


F43nd1r:
This currently fails due to an issue described here https://stackoverflow.com/questions/33136530/target-negation-not-working-in-spring-aop. I'm leaving it at this point today. Further investigation is needed, possibly need to raise the issue with spring aop team.

F43nd1r:
Now tracking upstream https://github.com/spring-projects/spring-framework/issues/27302

F43nd1r:
Workaround: Register your class extending ` GenericFilterBean` using ` @Bean` instead of ` @Component`

Issue #258, opened 2021-08-06
Enable/Disable Endpoint should show duration


WtfJoke:
Both if it were disable and if it is still active. Seeing something like: `durationMinutes: 180` makes for some simple math so the caller doesn't have to worry about when something was enabled (regardless of timezone).

Duration Is probably a poor name. `minutesActive` might be clearer. It would keep counting up if it hasn't been disabled. 

Feel free to ignore this idea though if it distracts from the endpoint's purpose.

_Originally posted by @checketts in https://github.com/codecentric/chaos-monkey-spring-boot/issues/246#issuecomment-890033899_

WtfJoke:
I like the idea. The feature is probably not that important but a nice addition.
It would make sense to have a good readable format for humans. I could imagine an output like "x hours x minutes x seconds"

MatthewBerk:
I am currently working on some other things at the moment but would still like to take on this issue!

MatthewBerk:
More things have come up in this past week, so I will get to working on this as soon as I can! Thank you for assigning it to me!

WtfJoke:
No need to appologize :)

PR #286, opened 2021-10-30
Endpoint minutes active

MatthewBerk:
**What**:
Added a time active and time inactive duration to the Enable/Disable endpoint timestamps.
Edited ChaosMonkeyDisabledDto and ChaosMonkeyEnabledDto to track duration Chaos Monkey was enabled and disabled for current enabled/disabled session.
Modified test cases for `enableChaosMonkey()` and `disableChaosMonkey()` methods in order to handle duration.

**Why**:
https://github.com/codecentric/chaos-monkey-spring-boot/issues/258

**How**:
Using java.time.Duration to calculate duration since ChaosMonkeyDisabledDto and ChaosMonkeyEnabledDto was created. Every time Chaos Monkey is enabled after being disabled and vice versa, a new instance of EnabledDto/DisabledDto is created since it represents current active and inactive session!

**Checklist**:

- [ ] Documentation added
      
- [x] Changelog updated
      
- [ ] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/ChaosMonkeyRestEndpoint.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/ChaosMonkeyRestEndpoint.java
@@ -42,6 +42,10 @@ public class ChaosMonkeyRestEndpoint {
 
   private final ChaosMonkeyScheduler scheduler;
 
+  private ChaosMonkeyEnabledDto chaosMonkeyEnabledDto;
+
+  private ChaosMonkeyDisabledDto chaosMonkeyDisabledDto;
+
   public ChaosMonkeyRestEndpoint(
       ChaosMonkeySettings chaosMonkeySettings,
       ChaosMonkeyRuntimeScope runtimeScope,
@@ -72,14 +76,26 @@ public class ChaosMonkeyRestEndpoint {
 
   @PostMapping("/enable")
   public ResponseEntity<ChaosMonkeyEnabledDto> enableChaosMonkey() {
-    this.chaosMonkeySettings.getChaosMonkeyProperties().setEnabled(true);
-    return ResponseEntity.ok().body(new ChaosMonkeyEnabledDto());
+    if (!this.chaosMonkeySettings.getChaosMonkeyProperties().isEnabled()
+        || chaosMonkeyEnabledDto == null) {
+      chaosMonkeyEnabledDto = new ChaosMonkeyEnabledDto();
+      this.chaosMonkeySettings.getChaosMonkeyProperties().setEnabled(true);
+    }
+    return ResponseEntity.ok().body(chaosMonkeyEnabledDto);
   }
 
   @PostMapping("/disable")
   public ResponseEntity<ChaosMonkeyDisabledDto> disableChaosMonkey() {
-    this.chaosMonkeySettings.getChaosMonkeyProperties().setEnabled(false);
-    return ResponseEntity.ok().body(new ChaosMonkeyDisabledDto());
+    if (this.chaosMonkeySettings.getChaosMonkeyProperties().isEnabled()
+        || chaosMonkeyDisabledDto == null) {
+      if (chaosMonkeyEnabledDto == null) {
+        chaosMonkeyDisabledDto = new ChaosMonkeyDisabledDto();
+      } else {
+        chaosMonkeyDisabledDto = new ChaosMonkeyDisabledDto(chaosMonkeyEnabledDto);
+      }
+      this.chaosMonkeySettings.getChaosMonkeyProperties().setEnabled(false);
+    }
+    return ResponseEntity.ok().body(chaosMonkeyDisabledDto);
   }
 
   @GetMapping
--- a/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
+++ b/chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
@@ -1,10 +1,52 @@
 package de.codecentric.spring.boot.chaos.monkey.endpoints.dto;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.time.Duration;
 import java.time.ZonedDateTime;
 import lombok.Data;
 
 @Data
 public class ChaosMonkeyDisabledDto {
   private final String status = "Chaos Monkey is disabled";
-  private ZonedDateTime disabledAt = ZonedDateTime.now();
+  private final ZonedDateTime disabledAt = ZonedDateTime.now();
+  @JsonIgnore private Duration timeActive;
+  @JsonIgnore private Duration timeInActive;
+  private String timeActiveFormatted;
+  private String timeInActiveFormatted;
+
+  public ChaosMonkeyDisabledDto() {
+    this.timeActive = Duration.ZERO;
+  }
+
+  public ChaosMonkeyDisabledDto(ChaosMonkeyEnabledDto chaosMonkeyEnabledDto) {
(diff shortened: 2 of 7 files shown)


WtfJoke:
Hey @MatthewBerk 
Currently im quiet busy. Takes probably a bit of a time until I'm able to review your PR. Sorry for the wait 🙇

MatthewBerk:
Understood @WtfJoke  👍

F43nd1r (commented):
I don't think the endpoints are the best place to store these timestamps. I'd expect the timestamps to be correct no matter what endpoint was used.

MatthewBerk:
> I don't think the endpoints are the best place to store these timestamps. I'd expect the timestamps to be correct no matter what endpoint was used.

So what you want is if enabled Chaos Monkey through Rest Endpoint, than say 1 minute later, enabled Chaos Monkey through JMX Endpoint, would want Chaos Monkey being active for 1 minute to be shown instead of 0 seconds (since the two endpoints holds their own ChaosMonkeyEnabledDto ChaosMonkeyDisabledDto objects).

 So both endpoint classes should refer to same ChaosMonkeyEnabledDto/ChaosMonkeyDisabledDto objects. 

Is this correct?

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
```diff
@@ -1,10 +1,52 @@
...
+
+    if (duration.toHours() > 0) {
+      return String.format(
+          "%d hours %02d minutes %02d seconds", duration.toHours(), minutesPart, secondsPart);
+    } else if (duration.toMinutes() > 0) {
+      return String.format("%d minutes %02d seconds", duration.toMinutes(), secondsPart);
+    } else {
+      return String.format("%d seconds", inSeconds);
+    }
+  }
```
WtfJoke, line 51: I would extract this method in a separate class so you can reuse it in both classes

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
```diff
@@ -1,10 +1,52 @@
...
 import java.time.ZonedDateTime;
 import lombok.Data;
 
 @Data
 public class ChaosMonkeyDisabledDto {
   private final String status = "Chaos Monkey is disabled";
-  private ZonedDateTime disabledAt = ZonedDateTime.now();
+  private final ZonedDateTime disabledAt = ZonedDateTime.now();
+  @JsonIgnore private Duration timeActive;
+  @JsonIgnore private Duration timeInActive;
```
WtfJoke, line 13: nitpicking here 🤓 If spotless allows it, put the `@JsonIgnore` above the field:
eg.: 
```java
@JsonIgnore
private Duration timeActive;
```

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/ChaosMonkeyRestEndpoint.java
```diff
@@ -42,6 +42,10 @@
 
   private final ChaosMonkeyScheduler scheduler;
 
+  private ChaosMonkeyEnabledDto chaosMonkeyEnabledDto;
+
+  private ChaosMonkeyDisabledDto chaosMonkeyDisabledDto;
```
WtfJoke, line 47: I think its better to store the state when chaos monkey was enabled in chaosmonkeysettings/chaosmonkeyproperties instead of here. Then you dont need to duplicate to Rest and JmxEndpoint

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
```diff
@@ -1,10 +1,52 @@
...
+    return timeInActive;
+  }
+
+  public String getTimeActiveFormatted() {
+    return formatDuration(timeActive);
+  }
+
+  public String getTimeInActiveFormatted() {
+    return formatDuration(getTimeInActive());
+  }
```
WtfJoke, line 36: These methods seems to be unused?

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/ChaosMonkeyRestEndpoint.java
```diff
@@ -72,14 +76,26 @@ public AssaultPropertiesUpdate getAssaultSettings() {
...
+    if (this.chaosMonkeySettings.getChaosMonkeyProperties().isEnabled()
+        || chaosMonkeyDisabledDto == null) {
+      if (chaosMonkeyEnabledDto == null) {
+        chaosMonkeyDisabledDto = new ChaosMonkeyDisabledDto();
+      } else {
+        chaosMonkeyDisabledDto = new ChaosMonkeyDisabledDto(chaosMonkeyEnabledDto);
+      }
+      this.chaosMonkeySettings.getChaosMonkeyProperties().setEnabled(false);
+    }
+    return ResponseEntity.ok().body(chaosMonkeyDisabledDto);
```
WtfJoke, line 98: I think as soon you store the enabled timestamp somewhere else, this could be simplified

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
```diff
@@ -1,10 +1,52 @@
...
+  private String timeActiveFormatted;
+  private String timeInActiveFormatted;
+
+  public ChaosMonkeyDisabledDto() {
+    this.timeActive = Duration.ZERO;
+  }
+
+  public ChaosMonkeyDisabledDto(ChaosMonkeyEnabledDto chaosMonkeyEnabledDto) {
+    this.timeActive = chaosMonkeyEnabledDto.getTimeActive();
+  }
```
WtfJoke, line 23: You could introduce an additional Constructor here which takes the timeactive as parameter.
Then you can call that from the other Constructors
eg. something like:
```java
  public ChaosMonkeyDisabledDto() {
    this(Duration.ZERO);
  }

  public ChaosMonkeyDisabledDto(ChaosMonkeyEnabledDto chaosMonkeyEnabledDto) {
    this(chaosMonkeyEnabledDto.getTimeActive());
  }
  public ChaosMonkeyDisabledDto(Duration timeActive){
    this.timeActive = timeActive;
    this.disabledAt = ZonedDateTime.now();
  }
```
There you can also set the disabledAt time stamp.

chaos-monkey-spring-boot/src/main/java/de/codecentric/spring/boot/chaos/monkey/endpoints/dto/ChaosMonkeyDisabledDto.java
```diff
@@ -1,10 +1,52 @@
...
 
 @Data
 public class ChaosMonkeyDisabledDto {
   private final String status = "Chaos Monkey is disabled";
-  private ZonedDateTime disabledAt = ZonedDateTime.now();
+  private final ZonedDateTime disabledAt = ZonedDateTime.now();
+  @JsonIgnore private Duration timeActive;
+  @JsonIgnore private Duration timeInActive;
+  private String timeActiveFormatted;
+  private String timeInActiveFormatted;
```
WtfJoke, line 15: They doesnt seem to be used as well. Probably not intentional

WtfJoke (changes requested):
Sorry for the long wait. I've finally gone through a first time. We've probably need another iteration later (as I havent checked the output yet), hope thats ok 😓

PR #289, opened 2021-11-16
Implemented the ability to use static factory methods for property cr…

codejnki:
**What**: Add an ability to create exception assaults that use exceptions that do not have a public constructor for creation.

**Why**: Some third-party libraries that throw custom exceptions created using a static factory method.  This allows for the creation of assaults that use this style of class.

**How**:

Added some additional properties on the exception argument class that allows for specifying a creation method to use to generate the exception class.

**Checklist**:

- [X] Documentation added
      
- [ ] Changelog updated
      
- [X] Tests added
      
- [X] Ready to be merged
      

Let me know if there are changes you'd like to see made.  This change greatly increases the style and types of exceptions that can be thrown.

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
@@ -292,6 +292,31 @@ Assault config has changed
 ----
 Assault config has changed
 ----
+====== Define custom Exceptions Without Constructors
+This is an example of configuration an exception which doesn't use a constructor, but instead uses a static factory method for creation.
+[source,json,subs="verbatim,attributes"]
+./chasomonkey/assaults - Request
+----
+{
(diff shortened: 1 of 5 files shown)


WtfJoke:
Hey @codejnki 
Thanks for your PR. 
My apologies for not taking a look yet. We will soon take a look at it.

WtfJoke (commented):
First of thanks for taking your time and creating a PR. 🙏 
I finally took a brief look at the PR.
However I am not sure if we should support that kind of behavior. I think its quite a bit of code/docs to maintain for a rather less used feature, so I tend to not support that.

I'm open for other opinions from my fellow maintainer friends such as @tuxdevelop @F43nd1r or @checketts

WtfJoke:
Just wanted to let you know that with the next release, the feature is available :)

PR #324, opened 2022-08-19
Add ConditionalOnMissingBean to assaults and advisors

F43nd1r:
**What**: Add ConditionalOnMissingBean to assaults and advisors, open `determineLatency` for extension

**Why**: Better extension support

**How**: Annotations.

**Checklist**:

- N/A Documentation added
      
- `Not sure if necessary to note` Changelog updated
      
- N/A Tests added
      
- [X] Ready to be merged


WtfJoke (approved):
I think we should note that in the changelog, that were now give our users the ability to overwrite our beans. I think you can call it a new feature. Code looks fine though (I hope we dont run in bean name clashes like in the past) :)

So after extending the changelog, feel free to merge 👍 
Thanks for taking care <3

PR #327, opened 2022-09-02
Add ApiDoc integration

F43nd1r:
**What**: Add springdoc-openapi integration

**Why**:

Show chaos monkey as a group in swagger ui

**How**:

configuration enabled by class and property

**Checklist**:

- [X] Documentation added
      
- [x] Changelog updated
      
- [N/A] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
@@ -2,6 +2,8 @@
 == Spring Boot Actuator Endpoints ==
 Chaos Monkey for Spring Boot offers you some built in endpoints exposed via https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints[JMX or HTTP]. This allows you to change configuration at runtime.
 
+https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html[Spring Boot Actuator] needs to be included in your project dependencies.
+
 '''
 
 === Enable chaosmonkey endpoint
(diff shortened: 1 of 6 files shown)


chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
```diff
@@ -2,6 +2,8 @@
 == Spring Boot Actuator Endpoints ==
 Chaos Monkey for Spring Boot offers you some built in endpoints exposed via https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints[JMX or HTTP]. This allows you to change configuration at runtime.
 
+https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html[Spring Boot Actuator] needs to be included in your project dependencies.
```
WtfJoke, line 5: This is a friendly hint, that you need Actuator when you want to use Actuator Endpoints right?

I was first a bit worried that all our users needs actuator even when they not use actuator endpoints :D

WtfJoke, line 5:
```suggestion
https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html[Spring Boot Actuator] needs to be included in your project dependencies if you want to use this feature.
```

Do you think this is a better or worser phrase?

chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
```diff
@@ -396,3 +398,25 @@ component=false
 |Is Chaos Monkey active or not
 |true or false
 |===
+
+=== Enable springdoc-openapi integration
+First enable wanted endpoints as shown above. Then include https://springdoc.org/[springdoc-openapi] in your dependencies and configure the following:
```
WtfJoke, line 403:
```suggestion
Enable wanted endpoints like described in XX(link to Spring Boot Actuator endpoints). Then include https://springdoc.org/[springdoc-openapi] in your dependencies and configure the following:
```

WtfJoke (commented):
Im quiet late to the party 🙈  but I've left some comments which I think makes it a bit easier to read (might be subjective).

chaos-monkey-docs/src/main/asciidoc/endpoints.adoc
F43nd1r: Yes, exactly. I just added this as a hint, we could mark it as such with a `Hint:` prefix?

WtfJoke: I've made an suggestion below. I think Hint: is not needed

PR #329, opened 2022-10-07
Upgrade to spring boot 2.7.4 and spring cloud 2021.0.4

WtfJoke:
**What**:
Update dependencies in pom.xml.

**Why**:
To keep 🐒  in shape 🏋️ 

**How**:
Checked new versions and use them.
Followed the [Upgrade guide from Spring Boot 2.6.x to 2.7.x](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7.0-Configuration-Changelog)

**Checklist**:

- [N/A] Documentation added
      
- [x] Changelog updated
      
- [N/A] Tests added
      
- [x] Ready to be merged


WtfJoke:
For some reason when replacing `@AutoConfiguration` with `@Configuration` the `@Autowired ApplicationEventPublisher` is null and I dont know why thats the case😿

WtfJoke:
It was because `@AutoConfiguration` uses proxyBeanMethods=false and when this is the case youre not allowed to call methods with `@Bean` annotation directly, which we do. So I replaced that and use dependency injection there as well

F43nd1r (approved):
LGTM 👍

PR #337, opened 2023-02-03
Extend repository class filter to also match proxied repositories.

### LINKED ISSUE #335: Repository watcher doesnt work
When enabling the repository watcher and a simple CrudRepository is used, there wont be any assaults.

## Expected Behavior
When enabling the repository watcher (`chaos.monkey.watcher.repository=true`) and a Repository is used which implements CrudRepository. Assaults should be triggered as soon as you use this repository.

## Current Behavior

When you use this repository, no assaults will be triggered.
## Steps to Reproduce

Demo-Repo: https://github.com/WtfJoke/chaoskotlindemo/tree/spring3
Start this app and use getMovie request from the file `requests.http`. There wont be any exception, although the exception assault (and the repository watcher) is enabled.
## Context (Environment)
it seems `org.springframework.data.jpa.repository.support.SimpleJpaRepository` is not considered in the current pointcuts.

F43nd1r:
Closes #335

**What**: 
Extend repository class filter to also match proxied repositories.
Add test to prevent regressions.

**Why**: 
See #335  - jdbc repositories were potentially not matched.

**How**: 
Pass on final proxies in jdk.proxy2.* in base filter to allow accepting proxied repositories.
Check directly `@Repository` annotated classes as done previously, but also check for annotations on interfaces in case of proxies.

**Checklist**: 

- N/A Documentation added
      
- [x] Changelog updated
      
- [X] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -5,6 +5,7 @@ Built with Spring Boot {spring-boot-version}
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
+* https://github.com/codecentric/chaos-monkey-spring-boot/pull/337[#337]: Jdbc repository watcher wasn't working in most cases.
 
 === Improvements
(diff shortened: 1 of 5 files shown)


WtfJoke (approved):
LGTM 🐦

PR #402, opened 2023-09-29
Always register outgoing watchers so they can be enabled later

### LINKED ISSUE #393: RestTemplate watcher not working when enabled via watchers actuator rest endpoint
Rest template watcher is not working when enabled via Watcher rest endpoint. In config, it shows enabled but assaults related to RestTemplate are not working. This is working fine when rest template watcher is enabled via application.properties. 

## Expected Behavior
If you are enabling RestTemplate watcher via POST watcher actuator endpoint then it should work fine.

## Current Behavior
When enabled via watcher actuator endpoint, it shows enabled in config but it is not working. The same when enabled via application.properties then it is working fine.
`chaos.monkey.watcher.rest-template=true`

## Steps to Reproduce
1. Enable RestTemplate watcher by sending a POST request to watcher actuator endpoint.
2. Verify watchers config by sending a GET request to watcher actuator endpoint. Here, RestTemplate will be shown as enabled.
3. Try any kind of assault related to RestTemplate. It will not work.
4. When RestTemplate watcher enabled by application.properties then assaults will work properly.

## Context (Environment)
Instead of enabling/disabling rest template watcher every time via application.properties, it will be help if it can be enabled/disabled via actuator endpoint.

F43nd1r:
**What**: Closes #393 

**Why**: outgoing watchers were previously only registered when the respective property was set, but there are other ways to enable those.

**How**: Remove ConditionalOnProperty

**Checklist**: 

- [ ] Documentation added
      
- [x] Changelog updated
      
- [ ] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -5,6 +5,7 @@ Built with Spring Boot {spring-boot-version}
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
+- https://github.com/codecentric/chaos-monkey-spring-boot/pull/402[#402] Always register outgoing watchers so they can be enabled later
 
 === Improvements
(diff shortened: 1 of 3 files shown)


WtfJoke (approved):
LGTM 🌿

PR #408, opened 2023-10-20
Support bean watching by class

### LINKED ISSUE #321: Add option to watch beans by type
Maybe this is already possible, or there is a reason why this is not implemented yet. As I couldn not find any information about this, I will create this issue.

## Expected Behavior
It should be possible to watch beans by type. It should also be possible to define a supertype, and all subtypes of it will also be watched.

## Current Behavior
For example if I want to enable chaos monkeys for all beans of type javax.sql.DataSource to simulate a total failure of the database.
Normally DataSource Beans have no @Component Annotation, so auto configuration does not work for them. I would have to configure the watcher by bean name. 

Using the bean name is not a very failsave solution in my opinion. Per default the bean name is derived from the method name annotated with @Bean. Simply renaming this method would break our chaos monkey configuration, and the bean is suddenly not watched anymore.

Configuring the Bean by type would solve this. As they type is still the same when the bean name changes.

## Possible Solution
I do not fully understand the code so far. But it might be fairly easy to implement this.

Extend the class ChaosMonkeyBeanPostProcessor
```java
if (watcherProperties.getBeans().contains(beanName) || isAssignable(bean)) {
    ...
}
...
private boolean isAsignable(Object bean) {
  if(watcherProperties.getBeanTypes().isEmpty) {
    return false;
  }

  Class<?> beanClass = bean.getClass();

  for(Class<?> requiredType : watcherProperties.getBeanTypes()) {
    if(requiredType.isAssignableFrom(beanClass)) {
      return true;
    }
  }

  return false;
}
```

And add a new configuration option

```chaos.monkey.watcher.beanTypes = javax.sql.DataSource,my.custom.Type```

F43nd1r:
**What**: Watch beans by type

**Why**: Closes #321 

**How**: Same as bean name

**Checklist**: 

- [X] Documentation added
      
- [X] Changelog updated
      
- [X] Tests added
      
- [X] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -10,11 +10,13 @@ Built with Spring Boot {spring-boot-version}
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
 
 === New Features
+ - https://github.com/codecentric/chaos-monkey-spring-boot/pull/408[#408] Watch beans by type
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
 
 === Contributors
 This release was only possible because of these great humans ❤️:
 
+ - https://github.com/F43nd1r[@F43nd1r]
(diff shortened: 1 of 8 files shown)


WtfJoke (approved):
LGTM, thank you @F43nd1r :) 👯

Issue #451, opened 2024-03-14
Counterintuitive behaviour while using WatchedCustomServices


eshaanganesh2:
Whenever we encounter any method, class, or package listed in _**WatchedCustomServices**_ and it satsifies the **_level_** criteria, an assault is launched. 

However, it's important to note that the level count increases each time we encounter a watched method, class, or package, regardless of whether it's present in the WatchedCustomServices list or not.

To simplify this, I propose that we only increase the level count `while encountering any watched method (or class or package) listed in WatchedCustomServices`. This adjustment would make the behavior more intuitive and ensure that the level count reflects the actual monitoring of the specified elements within the WatchedCustomServices list.

This can be better illustrated through an example.

## Current Behavior
Consider the below code which consists of a service layer implementation invoking 3 functions (which are defined in another class annotated with `@Component` )

```java
class serviceImpl(){ 

    @Autowired
    DemoApplicationComponent x;

    x.func1();
    x.func2();
    x.func3();
}
``` 
Let us assume the below configuration property has been set for WatchedCustomServices
`chaos.monkey.assaults.watchedCustomServices=com.example.demo.component.DemoApplicationComponent.func2`

and the watcher property for Component has been set to true
`chaos.monkey.watcher.component=true`

Let us also assume that a level count of 4 was set for this example
`chaos.monkey.assaults.level=4`
`chaos.monkey.assaults.deterministic=true`

When the request arrives at the serviceImpl class, the level count increases as follows:
```java
x.func1(); // level = 1 (Not present in WatchedCustomServices)
x.func2(); // level = 2 (Present in WatchedCustomServices) ; 2%4!=0, therefore no assault is launched
x.func3(); // level = 3 (Not present in WatchedCustomServices)
``` 

Upon making a 2nd request to the serviceImpl class, the values change as follows:
```java
x.func1(); // level = 4 (Not present in WatchedCustomServices)
x.func2(); // level = 5 (Present in WatchedCustomServices) ; 5%4!=0, therefore no assault is launched
x.func3(); // level = 6 (Not present in WatchedCustomServices)
``` 

Upon making a 3rd request to the serviceImpl class, the values change as follows:
```java
x.func1(); // level = 7 (Not present in WatchedCustomServices)
x.func2(); // level = 8 (Present in WatchedCustomServices) ; 8%4==0, therefore assault is launched!
x.func3(); // level = 9 (Not present in WatchedCustomServices)
``` 
[...]

PR #452, opened 2024-03-14
Increment level only for methods/classes/packages listed in WatchedCustomServices (instead of every watched method/class/package)

eshaanganesh2:
**What**: 
The level count increases every time we encounter a watched method, class, or package, regardless of whether it's present in the WatchedCustomServices list or not. Through this PR, I have made some changes to instead increase the level count only when a watched method (or class or package) is listed in WatchedCustomServices. 
**More details on this can be found at https://github.com/codecentric/chaos-monkey-spring-boot/issues/451** 

**Why**: 
These changes would help improve predictability wherein only methods/classes/packages listed within the WatchedCustomServices would increase level counts, as opposed to any and every watched method doing the same.

**How**: 
Changes have been made within the logic of `callChaosMonkey(ChaosTarget type, String simpleName)` function present in the `ChaosMonkeyRequestScope.java` file. 
A check has been implemented to verify if WatchedCustomServices  is active and if the exact same method is present in WatchedCustomServices . If yes, only then the level count is incremented.

**Checklist**: 

- [x] Documentation added
      
- [x] Changelog updated
      
- [ ] Tests added
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -1,13 +1,14 @@
 [[changes]]
-== Changes in {project-version}
+== Changes in 3.1.1
 
-Built with Spring Boot {spring-boot-version}
+Built with Spring Boot 3.2.3
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
 
 === Improvements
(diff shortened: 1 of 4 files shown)


denniseffing:
Hi @eshaanganesh2,
I'll close this PR in favor of #532. The code to determine whether or not an attack should be run was already a mess before you changed it and this PR made it slightly worse. However, we really appreciate the effort you put into the issue description and this PR. You are awesome and we credited you in the changelog accordingly. 🥳

PR #564, opened 2025-04-25
Fix repository watchers not correctly recognizing Spring Data repositories due to issues when matching their proxy package name

denniseffing:
**What**: We now use the `Proxy.isProxyClass` method instead of checking for package name prefixes.

**Why**: This is more reliable due to changing proxy package names.

Apparently, the package name of JDK proxies used by Spring Data repositories evolves over time. Testing with different JDK versions influenced the proxy package used by the repositories. Depending on the used version, we encountered the packages:
  - `com.sun.proxy` (JDK 8)
  - `jdk.proxy2` (JDK 11)
  - `jdk.proxy3` (JDK 16)
  - `jdk.proxy4` (JDK 17)

We therefore assume that the proxy package changes depending on the JDK version used.

HOWEVER: For some reason when executing the integration test `ChaosMonkeyRepositoryWatcherIntegrationTest` we encounter the package `jdk.proxy2` even though the tests were executed with JDK 21. This contradicts our theory that the proxy package changes depending on the JDK version only. We couldn't figure out why, but we didn't care anymore 🙃

**Checklist**: 

- [ ] Documentation added N/A
      
- [x] Changelog updated
      
- [ ] Tests added N/A (we couldn't figure out how to test this reliably)
      
- [x] Ready to be merged

### DIFF
--- a/chaos-monkey-docs/src/main/asciidoc/changes.adoc
+++ b/chaos-monkey-docs/src/main/asciidoc/changes.adoc
@@ -5,6 +5,7 @@ Built with Spring Boot {spring-boot-version}
 
 === Bug Fixes
 // - https://github.com/codecentric/chaos-monkey-spring-boot/pull/xxx[#xxx] Added example entry. Please don't remove.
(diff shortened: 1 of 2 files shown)


WtfJoke (approved):
LGTM

PR #659, opened 2026-09-14
Avoid proxying services with inherited final methods

### LINKED ISSUE #595: Unable to use Dependency Injection in Service class when having final method
Currently, we are trying to integrate Spring Boot Chaos Monkey library into our project. During the process, we wanted to create an experiment, which includes delayed response from one of our services.

The service under test is an abstract class annotated with @Service and @RequiredArgsConstructor and has a few classes, which are injected into it. To be more precise, we are testing a concrete implementation of the abstract class, but the method that we are testing is `public final` in the abstract class, therefore i am describing the abstract class.

When we add the dependency to Chaos Monkey library and we use `spring.profiles.active: chaos-monkey`, each dependency in the abstract class appear to be null. Therefore, it throws NPE, when we invoke the request. This happens **only when we activate the chaos-monkey profile** and is not the current production behaviour.

After digging down into the possible root cause for 2 days, I found out that the problem is actually caused by the **final** keyword in the method in the abstract class. When, I remove it works as expected, but it is pretty strange side effect. Moreover, changing the production code for the sake of integrating the library is not the desired approach, so i would be glad if you can assist with solving this issue and finding the root cause for it.

## Expected Behavior
When i activate the Spring Boot Chaos Monkey library, I do not expect to touch any of my production code and if i have not configured any assaults I want it to act as if there is not dependency to this library. In other words, I do not want to remove the `final` keyword from my code :)

## Current Behavior
Spring is not able to inject any dependency to the service that has final method and all of them appear to be null.

## Possible Solution
Removing `final` keyword from the method signature solves the issue for some reason.

## Steps to Reproduce
In my case, i am producing the issue with having an `abstract class` with `public final` method, which uses injected class in its logic. Apparently, all the injected classes in the class appear to be null (Spring is not able to inject them), so this could also be seen from debug console.
The issue is reproduced with the latest version 3.2.2. I also reproduced it with 3.1.0, 3.0.2, 3.1.2.
Our Spring Boot application is using Spring Boot 3.4.5.

## Context (Environment)
This issue becomes a deal breaker for integrating the library, because we are afraid of these side effects to our production code and we do not want test library integration to affect the behaviour of our newly developed features.

SamAyl:
**What**: Fix #595 by excluding service classes with non-private final instance methods from Chaos Monkey class-based proxying.

**Why**: An inherited final method executes on the CGLIB proxy rather than the injected target. Calling a dependency from that method can therefore throw a `NullPointerException` when the chaos-monkey profile is active.

**How**: Inspect the service class hierarchy for final instance methods, while leaving framework-generated proxies and other bean types alone. Add a Spring integration regression test with an abstract `@Service` base class and an injected dependency, plus a class-filter test and changelog entry. The two outgoing exception-assault integration tests now use local HTTP servers with the Spring-managed clients, preventing public website response headers from masking the injected exceptions.

**Validation**: Full `mvn verify -B` passes across all 10 reactor modules locally on Java 21. All 138 tests in the core module pass, with zero failures or errors. Spotless formatting/checks pass. GitHub’s Java/OS matrix still needs to run after maintainer workflow approval.

- [x] Documentation added (changelog)
- [x] Changelog updated
- [x] Tests added
- [x] Ready for review


SamAyl:
Hi @WtfJoke @denniseffing @jens-kaiser — could one of you please review this fix for #595 and approve the pending CI workflows when you have a chance?

The two failing exception-assault tests now use local HTTP servers instead of the public website. Full `mvn verify -B` passes locally on Java 21 across all 10 modules, including all 138 core module tests and Spotless checks. The GitHub Java/OS matrix is awaiting maintainer approval.

I’d appreciate your feedback and approval if the changes look good. Thank you!

